PR #29, opened 2017-07-02
Language: ? added, words changed.

comradekingu:
Quite a hefty oneliner to begin with, so diff can't be shown in online editor. This is the change:

s/Are you sure you want to delete "{folderName}" and all files inside. This operation can not be undone/Are you sure you want to delete "{folderName}" and all files in it. This can not be undone/


jospoortvliet:
:+1: makes sense

PR #31, opened 2017-07-03
Update README.md

jospoortvliet:
let's add a note about this issue pls... It is easy to work around but then ppl should know ;-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Once configured, the folders will show up in the home folder for each user in th
 
 ![folders](screenshots/folders.png)
 
-## A note about encryption
+## Notes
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -24,3 +24,4 @@ Once configured, the folders will show up in the home folder for each user in th
 ## A note about encryption
```
MariusBluem: So it’s not longer a note about encryption? 😅

jospoortvliet:
Whaddayathunk?

README.md
jospoortvliet: :see_no_evil:

MariusBluem:
:+1:

PR #32, opened 2017-07-03
Update info.xml

jospoortvliet:
add note on overwriting user folders?

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -9,7 +9,7 @@ Folders can be configured from *Group folders* in the admin settings.
 
 After a folder is created, the admin can give access to the folder to one or more groups, control their write/sharing permissions and assign a quota for the folder.
 
-Note: encrypting the contents of group folders is currently not supported.]]></description>
+Note: encrypting the contents of group folders is currently not supported. Group folders overwrite user folders with the same name so we recommend to create a unique name.]]></description>
 	<licence>AGPL</licence>
 	<author>Robin Appelman</author>
 	<version>1.1.0</version>
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -9,7 +9,7 @@ Folders can be configured from *Group folders* in the admin settings.
 
 After a folder is created, the admin can give access to the folder to one or more groups, control their write/sharing permissions and assign a quota for the folder.
 
-Note: encrypting the contents of group folders is currently not supported. Group folders overwrite user folders with the same name so we recommend to create a unique name.]]></description>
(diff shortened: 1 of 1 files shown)


jospoortvliet:
Well this should show up on https://apps.nextcloud.com I would say - and that shows the content of info.xml so... I think it should be shown.

MariusBluem:
A warning when creating group folders makes more sense. The info.xml should describe what an app does and not what does not work or what are restrictions. 

Or do you want to mention all issues in an info.xml 😅

jospoortvliet:
We could indeed do a warning when creating a folder. But in the mean time, these are two pretty big caveats people should be aware off so I think the app store should mention this. And as the app store content is created from info.xml...

MariusBluem:
> We could indeed do a warning when creating a folder.

Good idea 👍

jospoortvliet:
(there are of course more issues. Apparently versioning doesn't work, neither does Activities. Stuff to improve - but I don't think those have to be mentioned, only the most important stuff.

heldchen:
>  but I don't think those have to be mentioned, only the most important stuff.

lack of versioning is rather important IMO - we lost some data due to an accidental delete in a group folder today and had to recover from yesterdays backup... If I had known that versioning is not yet implemented, I'd have used a shared user folder for the critical stuff.

the current app description only mentions that encryption is missing, which can be misinterpreted that other standard features like Versioning and Trash are built in.

jospoortvliet:
@heldchen very good point, I added those too... I think this should be merged to notify users of these missing capabilities, for now - but I leave the final decision to @icewind1991  as it is his app ;-)

JK74:
@jospoortvliet @icewind1991 
In my believing: waste bin functionality and versioning are 'core' features of Nextcloud what cannot be missed. As such I would 'expect' that if an app does not follow or inherit these core functionality's that it should be very clearly defined in the app guidelines that are defined by Nextcloud; Maybe this can be discussed as a good ruling between APP developers and nextcloud?

I did used the app as well as I was excited about the shared quota etc as that is mandatory when using a shared folder like we do in our church; It resolved for me as well the issue that shared folders now have the issue that if they create a subdir/file that the subdir/file would then be still viewable by that user incl. all content in it when they 'leave' their assigned group that got shared this folder.  I'm now back at that level of shared folder security that I don't like to be in because I can't miss the waste bin and versioning. Is there any planning on waste-bin integration and or versioning?

Thnx again as always for keep on beeing so creative in new APP development!

tflidd:
Have the bugs been fixed or is this warning still valid?

jospoortvliet:
Yeah, this is now pretty much fixed so let's close...

PR #131, opened 2018-06-27
properly use indexes when querying folder size

### LINKED ISSUE #130: List all group folder on settings take a long time
Hello,

When we open the group folder overview on settings, then we must wait up to 4 minutes until we can edit the group folders. On furhter investigation we found the reason. 

nc_groupfolder

The function getAllFolders() in FolderManager.php makes an join to the filecache table on mariadb and our table has up to 15 million rows (8.5GB), therefore needs the query a long time.

Temporaly I have removed the "leftjoin" and "size".

```php
	public function getAllFolders() {
		$applicableMap = $this->getAllApplicable();

		$query = $this->connection->getQueryBuilder();

		$folderPath = $query->func()->concat($query->createNamedParameter('__groupfolders/'), 'folder_id');

		$query->select('folder_id', 'mount_point', 'quota')
			->from('group_folders', 'f');			

		$rows = $query->execute()->fetchAll();

		$folderMap = [];
		foreach ($rows as $row) {
			$id = $row['folder_id'];
			$folderMap[$id] = [
				'id' => $id,
				'mount_point' => $row['mount_point'],
				'groups' => isset($applicableMap[$id]) ? $applicableMap[$id] : [],
				'quota' => $row['quota'],
				'size' => $row['size'] ? $row['size'] : 0
			];
		}

		return $folderMap;
	}
```

icewind1991:
Fixes #130

@Gransi please verify if this fixes the problem for you

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -36,7 +36,8 @@ class Application extends App {
 
 		$container = $this->getContainer();
 		$container->registerService(FolderManager::class, function (IAppContainer $c) {
-			return new FolderManager($c->getServer()->getDatabaseConnection());
+			$rootStorageId = $c->getServer()->getRootFolder()->getMountPoint()->getNumericStorageId();
+			return new FolderManager($c->getServer()->getDatabaseConnection(), $rootStorageId);
 		});
 
 		$container->registerService(MountProvider::class, function (IAppContainer $c) {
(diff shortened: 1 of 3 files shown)


Gransi:
Today I replaced the two files on our NC installation and now it works.

Thanks you for the fast fix.

Gransi:
With the new fix I found a new bug. I can't make comments on folder on files.

The old version works.

PR #248, opened 2018-12-14
improve readabilty and point out caveat

TP75:
*Group folders* shall **not duplicate** the name of existing folders.


juliushaertl:
Closing as the caveat mentioned is fixed with https://github.com/nextcloud/groupfolders/pull/523

PR #315, opened 2019-02-12
No need for babel-polyfill, since it is already included with 16

juliushaertl:
Otherwise there will be complaints about babel-polyfill being loaded twice.


icewind1991:
already resolved with 836c07e95f446b3bdbb594528723418787ad9b53

Issue #267, opened 2018-12-18
TypeError: Argument 1 passed to OCA\\GroupFolders\\Versions\\VersionsBackend::getVersionsFolder() must be of the type int, string given, called in /var/www/nextcloud/apps/groupfolders/lib/Versions/VersionsBackend.php on line 154


mtippmann:
NC: 15 
Groupfolders: 2.0.2 

Seeing this in the Logfile from cron (that run's with php7.3 cli) - so it's maybe a PHP7.3 related error. 

```json
{
  "reqId": "bbg4JJ8VemzDijAzZ2uh",
  "level": 3,
  "time": "2018-12-18T21:15:05+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "cron",
  "method": "",
  "url": "--",
  "message": {
    "Exception": "TypeError",
    "Message": "Argument 1 passed to OCA\\GroupFolders\\Versions\\VersionsBackend::getVersionsFolder() must be of the type int, string given, called in /var/www/nextcloud/apps/
groupfolders/lib/Versions/VersionsBackend.php on line 154",
    "Code": 0,
    "Trace": [
      {
        "file": "/var/www/nextcloud/apps/groupfolders/lib/Versions/VersionsBackend.php",
        "line": 154,
        "function": "getVersionsFolder",
        "class": "OCA\\GroupFolders\\Versions\\VersionsBackend",
        "type": "->",
        "args": [
          "1"
        ]
      },
      {
        "file": "/var/www/nextcloud/apps/groupfolders/lib/Versions/GroupVersionsExpireManager.php",
        "line": 53,
        "function": "getAllVersionedFiles",
        "class": "OCA\\GroupFolders\\Versions\\VersionsBackend",
        "type": "->",
        "args": [
          {
            "id": "1",
            "mount_point": "StuRa",
            "groups": {
              "StuRa": "31"
            },
            "quota": "-3",
            "size": 0
          }
        ]
      },
      {
        "file": "/var/www/nextcloud/apps/groupfolders/lib/BackgroundJob/ExpireGroupVersions.php",
        "line": 40,
        "function": "expireAll",
        "class": "OCA\\GroupFolders\\Versions\\GroupVersionsExpireManager",
        "type": "->",
        "args": []
      },
      {
        "file": "/var/www/nextcloud/lib/private/BackgroundJob/Job.php",
        "line": 61,
        "function": "run",
[...]

PR #326, opened 2019-02-18
Followup patch fixing null permissions bug #267

Sprinterfreak:


icewind1991:
Thanks for the quick fix

PpaiSanchez:
Thanks icewind1991, the code looks good now.

PR #350, opened 2019-03-14
Explicitly cast timestamp

alexandernst:
This avoids passing an empty string instead of an integer when used with S3 backend as primary storage.


icewind1991:
thanks for the fix

PR #454, opened 2019-05-24
Fix if user in session is null

xf-:
Is it is a shared link no user session exists.


icewind1991:
Thanks for the fix, but I already had this sitting around uncommited (f02934a198df9ef07dafc4ad28b0035621ed7431)

Thanks anyway

xf-:
@icewind1991 do you also have a fix for cron? I got a lot of entries because you call a function accessing the userSession. I fixed it with `\OC::$CLI`.

PR #507, opened 2019-07-09
Use new sidebar api to register groupfolders section

juliushaertl:
Requires https://github.com/nextcloud/server/pull/16312

### DIFF
--- a/src/files.js
+++ b/src/files.js
@@ -20,6 +20,9 @@
  *
  */
 
+__webpack_nonce__ = btoa(OC.requestToken);
+__webpack_public_path__ = OC.linkTo('groupfolders', 'build/');
+
 (function(OC, OCA) {
 	OC.Plugins.register('OCA.Files.App', {
 		attach: () => {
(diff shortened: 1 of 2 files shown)


src/files.js
```diff
@@ -30,30 +33,15 @@
...
-	OC.Plugins.register('OCA.Sharing.ShareTabView', ShareTabPlugin);
 })(OC, OCA);
+
+window.addEventListener('DOMContentLoaded', () => {
+	import(/*c webpackChunkName: "sharing" */'./SharingSidebarApp').then((Module) => {
+		OCA.Sharing.ShareTabSections.registerSection((el, fileInfo) => {
+			if (fileInfo.mountType !== 'group') {
+				return
+			}
+			return Module.default
```
skjnldsv: interesting, dynamic import needs us to specify `defaukt` ?

juliushaertl: yes, it will return an object with all the exports

skjnldsv:
Shall we merge?

juliushaertl:
Yes, we need to make sure to have a version release just for 18 then

PR #519, opened 2019-07-15
Make sure we always trim the first path separator

juliushaertl:
Otherwise we end up with a rules cache with duplicate entries for the same path
that might not contain ACL rules


juliushaertl:
3.x doesn't have the ACL feature so no need to backport to that.

Issue #400, opened 2019-04-25
advanced permissions: who is allowed to edit the permissions?


schiessle:
After a short try of the new advanced permissions it looks like only admins can change the advanced group folder permissions. But what happens if no admin has access to the group folder, e.g. because non of the group members is an admin?

I think we should allow also group admins, to manipulate the permissions of their group folders.

wiswedel:
Duplicate of #366

wiswedel:
> add a column to the group folder settings where the admin configure the group folders to select the users which are allowed to manipulate the permissions

Keep in mind that the pickable users would need to be members of the previously selected groups.

We could substitute the "Advanced permissions" checkbox with that user selection.

@icewind1991 any opinion about this? This is a real production-blocker for larger organizations.

juliushaertl:
@schiessle @wiswedel Wouldn't it make more sense to give allow groups to either change the ACL or not? That way the actual user selection can also be managed by external user backends?

UI wise I have something like this in mind:
![image](https://user-images.githubusercontent.com/3404133/61200723-b28e4a00-a6e2-11e9-9d55-5c6d70767448.png)

wiswedel:
> @schiessle @wiswedel Wouldn't it make more sense to give allow groups to either change the ACL or not? That way the actual user selection can also be managed by external user backends?
> 
> UI wise I have something like this in mind:
> ![image](https://user-images.githubusercontent.com/3404133/61200723-b28e4a00-a6e2-11e9-9d55-5c6d70767448.png)

Totally agree. Best approach from my pov

juliushaertl:
Implementation is in https://github.com/nextcloud/groupfolders/pull/522, testing is very welcome

nursoda:
This is related to https://github.com/nextcloud/groupfolders/issues/400

juliushaertl:
@nursoda This is #400 :grin:

nursoda:
> @nursoda This is #400 grin

Sorry, I mixed both posts up, correct now, please delete off-topic

PR #522, opened 2019-07-15
Allow groups to manage ACL permissions

juliushaertl:
Implements #400 

This adds another permission setting to the groups of a groupfolder, so the admin can specify, who is able to change the ACL rules of a group folder. The setting is stored in a separate database field so we keep the regular permissions field clean to what is defined as OCP constants.

ToDo:
- [x] Add OCC command to change manage acl permissions


wiswedel:
Thanks @juliushaertl for that quick implementation.
I gave it a test, but noticed 2 critical things:
1. Users from newly entitled groups now can set permissions, however, those are only visible to the setter and also don't seem to apply to the affected right recipient.
2. Managing needs to be granted in the actual ACL setter as well. That's the whole point, to entitle specific users to manage deeper access rights to other specific people or groups. Otherwise you'd need a whole lot of extra groups only to manage the entitled users.

juliushaertl:
Thanks for giving it a run @wiswedel 

> Users from newly entitled groups now can set permissions, however, those are only visible to the setter and also don't seem to apply to the affected right recipient.

This is working as expected here. A newly added group doesn't have permission to manage the ACL and after enabling the checkbox the UI shows the option to change the permissions. Did you test with a regular user or an admin user? I missed to adjust the check for admin users, which might be what caused this for you ... fix incoming.

> Managing needs to be granted in the actual ACL setter as well. That's the whole point, to entitle specific users to manage deeper access rights to other specific people or groups. Otherwise you'd need a whole lot of extra groups only to manage the entitled users.

This was never mentioned in the original issue. From my understanding the use case was always to have a subgroup of the users who can access the group folder that can then manage the advanced permissions. Having this on a per subfolder level is much more complex on the implementation side. I partly get the use case for such a feature but it feels that this will make ACL rules even more complicated than they are already. Is that really a required part of the implementation? What is the issue with just having an "groupfolder-admin" group on a per-groupfolder level?

wiswedel:
> What is the issue with just having an "groupfolder-admin" group on a per-groupfolder level?

We'd need special sub-groupfolder-admins for each folder:
- groupfolder-admin-marketing
- groupfolder-admin-sales
- groupfolder-admin-engineering
...

That can become quite a list.

If we only had a group _groupfolder-admin_, each member of that group would have access to all administerable other group folders.

wiswedel:
> This was never mentioned in the original issue. From my understanding the use case was always to have a subgroup of the users who can access the group folder that can then manage the advanced permissions. 

Actually it was:

> [...] add a column to the group folder settings where the admin configure the group folders to select the users which are allowed to manipulate the permissions. Group admins can do other things like creating new users in their group which might be to much, therefore a dedicated setting might make more sense.

> Keep in mind that the pickable users would need to be members of the previously selected groups.
We could substitute the "Advanced permissions" checkbox with that user selection.

juliushaertl:
> add a column to the group folder settings where the admin configure the group folders to select the users which are allowed to manipulate the permissions. Group admins can do other things like creating new users in their group which might be to much, therefore a dedicated setting might make more sense.

> Keep in mind that the pickable users would need to be members of the previously selected groups.
We could substitute the "Advanced permissions" checkbox with that user selection.

Hm, maybe I got that wrong then, but as far as I understand it, that was still on a per groupfolder level but having the possibility to choose individual users, not having indivdual acl manage permissions for subdirectories of group folders.

wiswedel:
> Hm, maybe I got that wrong then, but as far as I understand it, that was still on a per groupfolder level but having the possibility to choose individual users, not having indivdual acl manage permissions for subdirectories of group folders.

Per groupfolder level would be sufficient, yes. The individual approach would be the golden bonus level 😉

juliushaertl:
Ok, let me see what I can do there then :wink:

wiswedel:
@juliushaertl Yesss. This looks nice. And the user selection menu is so snappy ❤️

So the bonus level gets put to the backlog?

juliushaertl:
> So the bonus level gets put to the backlog?

Yes, let's keep that as an possible enhancement for later.

Ready for review.

wiswedel:
@juliushaertl Looks great and works like a charm. Thanks for your work. This makes the app way more production ready for large environments.

wiswedel:
@rullzer @icewind1991 Can you please give @juliushaertl 's changes a review too?
Thx.

wiswedel:
resolves #561

wiswedel:
@rullzer @icewind1991 I really need this for two major customers. So without feedback from your side, I'll have to merge this unrevisioned to get a release on the way.
Thx.

miniorange-sudhanshu:
how to get a nextcloud admin account ?

Issue #18, opened 2017-06-09
Unexpected behavior when creating a group folder with the same name as an existing user folder.


dan-nkl:
Version: latest master...?

1. Create folder with content and subfolders
2. Make the folder to a groupfolder
3. ???
4. Content gone!

The content of the folder should be moved into the "created" groupfolder.

jospoortvliet:
How did you do step 2? Or is this a feature request, that an admin can turn a folder into a group folder?

jospoortvliet:
There's a weird overlap with existing folders. Take this scenario:

* Go to the Files app and create a folder named `superawesometest`
* Put a few files in it.
* Go to the Group Folders admin and create a folder named `superawesometest`
* Add the Admin group to the access rights for `superawesometest`
* Go back to Files and observe the new Group Folder has overwritten the existing `superawesometest` and it is empty. Where are my files?!!?!
* Go to the Group Folders admin and remove Admin group access to the folder `superawesometest`
* Go back to Files and observe the new Group Folder has disappeared and the existing `superawesometest` is back with its original content!
* Note that this overwriting of existing folders with (empty) Group Folders also happens with users - if you create a group folder with the same name as a folder one of your users has, they will lose access to that folder and all its content.

I think Group Folders should act like shares - if somebody shares a folder with the same name as an existing one with you, it should show up besides the existing one, not overwrite it.

heldchen:
ran into this issue as well, but had an additional side effect of ending up with multiple shared folderswith the same name

1. create normal folder, share with group
2. create group folder with same name

- users now see *two* shared folders
- deleting the Shared Folder results in an error, share not removed
- deleting the Group Folder removes all content but restores the Shared Folder

the workaround for the duplicated shares is to first save the folder contents, remove the Group Folder, then remove the Shared Folder, then recreate the Group Folder.

icewind1991:
the problem with trying to do any kind of conflict resolution on creating group folders (whether it is moving the files into the group folder, or even just showing a warning), is that it would have to check for conflicts for every user that has the folder which is much to expensive to do on large instances.

wiswedel:
Fixed by #523

kesselb:
@wiswedel should we reopen? #523 has been reverted.

wiswedel:
> @wiswedel should we reopen? #523 has been reverted.

totally I guess :-(

wiswedel:
ref: https://github.com/nextcloud/groupfolders/pull/647

[...]

PR #523, opened 2019-07-16
Check for naming conflicts before returning the user mounts

juliushaertl:
This PR makes sure we check and rename existing files/shares to a unique name before the groupfolder mount is added to the user mounts.

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -85,7 +85,19 @@ class MountProvider implements IMountProvider {
 	public function getMountsForUser(IUser $user, IStorageFactory $loader) {
 		$folders = $this->getFoldersForUser($user);
 
-		return array_map(function ($folder) use ($user, $loader) {
+		/** @var Folder $folder */
+		$userHome = \OC::$server->getRootFolder()->getUserFolder($user->getUID());
+		return array_map(function ($folder) use ($user, $loader, $userHome) {
+			// check for existing files in the user home and rename them if needed
+			$folderName = $folder['mount_point'];
+			$i = 1;
+			while($userHome->nodeExists($folderName)) {
+				$folderName = $folder['mount_point'] . ' (' . $i++ . ')';
+			}
+			if ($folderName !== $folder['mount_point']) {
+				$userHome->get($folder['mount_point'])->move($userHome->getPath() . '/' . $folderName);
+			}
+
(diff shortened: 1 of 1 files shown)


juliushaertl:
@rullzer Any feedback on the approach? I think performance wise this should be fine since it basically adds only one check for existing file by default.

lib/Mount/MountProvider.php
```diff
@@ -85,7 +85,19 @@ public function getFoldersForUser(IUser $user) {
 	public function getMountsForUser(IUser $user, IStorageFactory $loader) {
 		$folders = $this->getFoldersForUser($user);
 
-		return array_map(function ($folder) use ($user, $loader) {
+		/** @var Folder $folder */
+		$userHome = \OC::$server->getRootFolder()->getUserFolder($user->getUID());
```
rullzer: Doesn't this trigger cyclic things?

juliushaertl: Didn't seem so when testing, but indeed this seems like it could cause issues. I'll dig a bit deeper tomorrow.

juliushaertl: Looks fine, as the mountpoints are only setup once for the user. So when getRootFolder is called, initMountPoints will not execute again.

https://github.com/nextcloud/server/blob/1b85ef4bf2a18f7ee7bca7bcd3d7702362125d2e/lib/private/Files/Filesystem.php#L403

The issue I see would be that mount providers that are registered after the group folder provider don't have their mountpoints added to the user root at that point. For sharing mounts this is handled by manually forcing that mount provider to be the last mounted. https://github.com/nextcloud/server/blob/3ad6084891b88b3b7ef784bbf297b0a57b282d77/lib/private/Files/Config/MountProviderCollection.php#L93-L101

But I assume we should only rename regular users files anyway, so that shouldn't be an issue?

rullzer: Ok lets just try it then...

juliushaertl:
@icewind1991 Would be great if you could have a look as well

wiswedel:
Fixes https://github.com/nextcloud/groupfolders/issues/18

rullzer:
Ok so I was right. It does make boom. Due to cyclic nitialization. And actually causes: https://github.com/nextcloud/groupfolders/issues/597

PR #542, opened 2019-07-30
Return proper values from AclStorageWrapper

juliushaertl:
Otherwise the methods only return a boolean value instead of the actual result.


icewind1991:
:see_no_evil: 

(I think I was confused with js where you can use this trick)

rullzer:
@icewind1991 js confuses us all

juliushaertl:
Also took me a while to look at until I god why that was not working properly :wink:

wiswedel:
fixes #551

PR #543, opened 2019-07-30
fixed api routes to doc spec, resolves #499

### LINKED ISSUE #499: Fresh install has incorrect OCS endpoints
I've just recently update my server to NextCloud 16.0.1 along with installing GroupFolders to try and test some integration.
when I try and make a test call
`curl -u $USER:$PASS -X GET -H "OCS-APIRequest: true" \
"$SERVER/nextcloud/ocs/v1.php/apps/groupfolders/folders" `
it responds saying that it doesn't recognize the query
`<?xml version="1.0"?>
<ocs>
 <meta>
  <status>failure</status>
  <statuscode>998</statuscode>
  <message>Invalid query, please check the syntax. API specifications are here: http://www.freedesktop.org/wiki/Specifications/open-collaboration-services.
</message>
 </meta>
 <data/>
</ocs>`

I did a bit of digging into which routes are being registered and it seems that a prefix is being incorrectly applied to the requested route.
requested routes: "/ocs/apps/groupfolders/folders", "/ocsapp/apps/groupfolders/folders"
known route: "/apps/groupfolders/folders"

As a workaround, I found that changing line 2 of apps/groupfolders/appinfo/routes.php to `return ['ocs' => [` will get it working.

vardrop:


juliushaertl:
Thanks for your pull request, unfortunately just updating the routes won't work as they are already in use in various places like the frontend. Docs have been adjusted to the actual rules.

PR #552, opened 2019-08-08
Don't use arrow function in files app integration for IE11 compatibility

juliushaertl:


juliushaertl:
Hmm, right, maybe we we are missing the @babel/plugin-transform-arrow-functions
plugin. Let me have another look.

PR #600, opened 2019-09-11
Check if the parent folder is updatable when moving

juliushaertl:
When moving folders we should also check if the parent folder is actually updatable. Otherwise users could mess with the folder structure.

Try moving FolderA into FolderB with the following ACL applied:
- Groupfolder/  READ 
- Groupfolder/FolderA/  READ WRITE UPDATE DELETE
- Groupfolder/FolderB/  READ WRITE UPDATE DELETE

### DIFF
--- a/lib/ACL/ACLStorageWrapper.php
+++ b/lib/ACL/ACLStorageWrapper.php
@@ -88,8 +88,11 @@ class ACLStorageWrapper extends Wrapper {
 			}
 		}
 		$permissions = $this->file_exists($path2) ? Constants::PERMISSION_UPDATE : Constants::PERMISSION_CREATE;
-		$checkParentPermissions = !$this->is_dir($path1) || $this->checkPermissions(dirname($path1), Constants::PERMISSION_UPDATE);
-		return $checkParentPermissions &&
+		$sourceParent = dirname($path1);
+		if ($sourceParent === '.') {
+			$sourceParent = '';
+		}
+		return $this->checkPermissions($sourceParent, Constants::PERMISSION_DELETE) &&
 			$this->checkPermissions($path1, Constants::PERMISSION_UPDATE & Constants::PERMISSION_READ) &&
 			$this->checkPermissions($path2, $permissions) &&
 			parent::rename($path1, $path2);
--- a/lib/ACL/ACLStorageWrapper.php
+++ b/lib/ACL/ACLStorageWrapper.php
@@ -88,7 +88,9 @@ class ACLStorageWrapper extends Wrapper {
 			}
 		}
 		$permissions = $this->file_exists($path2) ? Constants::PERMISSION_UPDATE : Constants::PERMISSION_CREATE;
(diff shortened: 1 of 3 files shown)


lib/ACL/ACLStorageWrapper.php
```diff
@@ -88,7 +88,9 @@ public function rename($path1, $path2) {
 			}
 		}
 		$permissions = $this->file_exists($path2) ? Constants::PERMISSION_UPDATE : Constants::PERMISSION_CREATE;
-		return $this->checkPermissions($path1, Constants::PERMISSION_UPDATE & Constants::PERMISSION_READ) and
+		$checkParentPermissions = !$this->is_dir($path1) or $this->checkPermissions(dirname($path1), Constants::PERMISSION_UPDATE);
+		return $checkParentPermissions and
```
nickvergessen:
```suggestion
		return $checkParentPermissions &&
```

lib/ACL/ACLStorageWrapper.php
```diff
@@ -88,7 +88,9 @@ public function rename($path1, $path2) {
...
+			$this->checkPermissions($path1, Constants::PERMISSION_UPDATE & Constants::PERMISSION_READ) and
```
nickvergessen:
```suggestion
			$this->checkPermissions($path1, Constants::PERMISSION_UPDATE & Constants::PERMISSION_READ) &&
```

lib/ACL/ACLStorageWrapper.php
```diff
@@ -88,7 +88,9 @@ public function rename($path1, $path2) {
...
 			$this->checkPermissions($path2, $permissions) and
```
nickvergessen:
```suggestion
			$this->checkPermissions($path2, $permissions) &&
```

juliushaertl:
> They are not equal, and in fact and means the options are all tested, so parent::rename() is actually called.

Ah thanks, I wasn't aware of that :confounded: I made sure that `&&` is now used all over the place, since other operations were also affected by this.

lib/ACL/ACLStorageWrapper.php
```diff
@@ -88,7 +88,9 @@ public function rename($path1, $path2) {
 			}
 		}
 		$permissions = $this->file_exists($path2) ? Constants::PERMISSION_UPDATE : Constants::PERMISSION_CREATE;
-		return $this->checkPermissions($path1, Constants::PERMISSION_UPDATE & Constants::PERMISSION_READ) and
+		$checkParentPermissions = !$this->is_dir($path1) or $this->checkPermissions(dirname($path1), Constants::PERMISSION_UPDATE);
```
icewind1991: Why only do this check for folders?

skjnldsv:
@nickvergessen @icewind1991 can any of you help on this pr?
This is high priority and Julius is in vacations :pray:

skjnldsv:
Thanks @icewind1991 :heart:

Issue #271, opened 2019-01-03
Sharing via api not possible


pebosi:
Created a new group folder "Groupfolder" added a folder "Test" and tried to share via shares api:

curl --request POST   --url 'https://box/ocs/v2.php/apps/files_sharing/api/v1/shares?format=json'   --header 'authorization: Basic YWRtaW46blhngtTW9CSHEtb0dUZ0gtSkplZ0gtRGlDMmI='   --header 'content-type: application/x-www-form-urlencoded'   --header 'ocs-apirequest: true'   --data 'shareType=3&path=%2FGroupfolder%2FTest'

Resulted in Error 403.
Sharing the folder via Web-interface works as expected.

Recently updated to Nextcloud 15 and newest version of groupfolders. With Nextcloud 14, 13 it worked.

pebosi:
The base64 string above is for sure not real. Tried differnt users with permissions to the group folder.

JanUrlaub:
Same here. is there some new infos?

ricain:
Same issue here but with APIv1
#nextcloud/ios/issues/795

jospoortvliet:
Also mentioned here: https://github.com/nextcloud/server/issues/14189 - with some more example.

juliushaertl:
From a quick look it seems that https://github.com/nextcloud/server/blob/b557f52c2290379b99838f86c7ab90892b955b41/lib/private/Share20/Manager.php#L296 is causing the issues here. The original node doesn't have an owner for files in a group folder. I'm not sure yet, why it does work in the web ui then, but maybe the mounts are set up differently with an active user session.

ricain:
so this is a [nextcloud/server](https://github.com/nextcloud/server) bug and not a [nextcloud/groupfolder](https://github.com/nextcloud/groupfolders) issue ?

radoslavius:
I found the reason and solution.

Reason:
When attempting to create a share, the nextcloud does not detect a correctly logged user. As @juliushaertl wrote: same code `$share->getNode()->getOwner()->getUID()` in `lib/private/Share20/Manager.php` returns empty string instead of username.

Solution:
So if you use the guzzle library, just set the 'cookies' parameter to true and call `GET /shares` (for example) before calling `POST /shares`. This way the user logs in and the share is created correctly.

Tested on version 16.0.4

juliushaertl:
https://github.com/nextcloud/groupfolders/issues/271 should fix this

fuco809:
[...]

PR #634, opened 2019-10-09
Return proper owner for storage

juliushaertl:
When sharing a file from a groupfolder with the ocs api without an active session the trash and version backend are registered before the current user session has been initialized though the login. This leads to https://github.com/nextcloud/server/blob/4560bf939b32c4e07550bc1280827dcebc4eada1/lib/private/Files/Storage/Common.php#L406-L410 being called and caching false as the owner for the group folder storage.

Now when sharing a file it is assumed that there is always an owner (which will be fixed with https://github.com/nextcloud/server/pull/17489) but I assume it would be safer to make sure the current user is always assumed as the owner in case it is checked in other places as well.

Fixes https://github.com/nextcloud/groupfolders/issues/271

### DIFF
--- a/lib/Mount/GroupFolderStorage.php
+++ b/lib/Mount/GroupFolderStorage.php
@@ -24,6 +24,7 @@ namespace OCA\GroupFolders\Mount;
 
 use OC\Files\Storage\Wrapper\Quota;
 use OCP\Files\Cache\ICacheEntry;
+use OCP\IUserSession;
 
 class GroupFolderStorage extends Quota {
 	/** @var int */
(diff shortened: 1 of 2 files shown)


juliushaertl:
Another idea would be to adjust the Common storage to always recheck if getUser returns an actual user, maybe @rullzer or @icewind1991 can tell if that would be the better approach or if there are any issues expected with that.

lib/Mount/GroupFolderStorage.php
```diff
@@ -44,6 +44,10 @@ public function getFolderId() {
 		return $this->folderId;
 	}
 
+	public function getOwner($path) {
+		return \OC_User::getUser();
```
icewind1991: please inject a `UserSession` instead of using static methods

lib/Mount/GroupFolderStorage.php
```diff
@@ -32,18 +33,27 @@ class GroupFolderStorage extends Quota {
...
+		$this->userSession = $parameters['userSession'];
 	}
 
 	public function getFolderId() {
 		return $this->folderId;
 	}
 
+	public function getOwner($path) {
+		$user = $this->userSession->getUser();
+		return $user !== null ? $user->getUID() : false;
```
juliushaertl: false is returned here to keep the behavior aligned with the current common implementation. 

When false is returned getOwner on the Node object returns an IUser object with empty uid, which is assumed at some places where getOwner is not checked for null. While not nice, this avoids breaking other existing stuff that calls getOwner

juliushaertl:
:ping_pong:

PR #662, opened 2019-11-04
Search for users by displayname

juliushaertl:
To be reviewed together https://github.com/nextcloud/server/pull/17800 or with LDAP as a backend


juliushaertl:
Different approach that doesn't rely on the individual group backend search.

PR #688, opened 2019-12-05
Check for ACL list in trash backend

### LINKED ISSUE #610: All users can view any deleted file regardless of rights given to subfolders
We have group folders for each department in our organisation. Each group folder has the respective department user group added with write/share/delete rights in the Group Folder settings.
Then sub-folders under the group folders have differing advanced permissions starting with deny read for the whole group & then adding other permissions for those that need them.

The problem that we have found and it is a potentially very damaging one for sensitive information is, anyone that has any rights for the original group folder can view ANY file that has been deleted in ANY sub folder whether they have read rights or not in the deleted files.
Is there anyway to prevent this or am I using the advanced permissions incorrectly?

juliushaertl:
Fixes #610 

- [x] Check ACL when listing the trash
- [x] Check ACL when accessing the node for the trash item (restore/delete)
- [x] Check ACL when accessing the preview

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -135,6 +135,7 @@ class TrashBackend implements ITrashBackend {
 		if ($storage->instanceOfStorage(GroupFolderStorage::class) && $storage->isDeletable($internalPath)) {
 			/** @var GroupFolderStorage|Jail $storage */
 			$name = basename($internalPath);
+			$fileEntry = $storage->getCache()->get($internalPath);
 			$folderId = $storage->getFolderId();
 			$trashFolder = $this->getTrashFolder($folderId);
 			$trashStorage = $trashFolder->getStorage();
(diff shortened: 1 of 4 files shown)


lib/Trash/TrashBackend.php
```diff
@@ -243,7 +262,8 @@ public function getTrashNodeById(IUser $user, int $fileId) {
 			$absolutePath = $this->appFolder->getMountPoint()->getMountPoint() . $path;
 			$relativePath = $trashFolder->getRelativePath($absolutePath);
 			list(, $folderId) = explode('/', $relativePath);
-			if ($this->userHasAccessToFolder($user, (int)$folderId)) {
+			// FIXME doesn't work yet since we do not have the original path
```
juliushaertl: @icewind1991 I could need some help here, i fail to find the connection between the appdata fileId and the entry in `oc_group_folder_trash` so I'm not sure how I could obtain the original path inside of this method.

icewind1991: I guess you'll have to change the explode above in

```php
list(, $folderId, $nameAndTime) = explode('/', $relativePath);
```

then split the `$nameAndTime` (on the last `.`) and get search in the trash folder based on that.

"Going the wrong way" like this is awkward, which is why the trash api returns the `Node` instead of a `TrashItem` in the first place :)

juliushaertl: @icewind1991 The main issue with this is that we cannot obtain the original path just from that as the name+time are not really unique identifiers. I've pushed a commit to add the fileid to the oc_group_folders_trash table so we can properly fetch the original path to check against acl rules. Let me know what you think about that.

juliushaertl:
I'll add them in a follow up.

PR #761, opened 2020-02-12
Show inherited acls

nenadalm:
This pr adds inherited advanced rules into the sidebar together with rules for current file. This allows seeing all advanced rules for the file with possibility of changing any rule (rule for current file will be changed in case inherited rule is shown).

~~When rule is removed in the ui, it is removed from the list and if it was inherited, it will be shown there again after reopening (maybe this should be changed somehow?).~~

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -140,21 +140,20 @@ client.addFileInfoParser(function(response) {
 		data.aclCanManage = !!aclCanManage;
 	}
 
-	var acls = props[ACL_PROPERTIES.PROPERTY_ACL_LIST];
-	var inheritedAcls = props[ACL_PROPERTIES.PROPERTY_INHERITED_ACL_LIST];
+	var acls = props[ACL_PROPERTIES.PROPERTY_ACL_LIST] || [];
+	var inheritedAcls = props[ACL_PROPERTIES.PROPERTY_INHERITED_ACL_LIST] || [];
 
-	if (!_.isUndefined(acls)) {
-		data.acl = parseAclList(acls);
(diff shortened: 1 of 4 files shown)


janpekar:
I recommend do not allow to delete permission, that is inherited.

nenadalm:
> I recommend do not allow to delete permission, that is inherited.

I removed the confusing behaviour and now inherited rules cannot be removed. In case they're changed and changed rule is removed, inherited rule takes it's place in the ui.

juliushaertl:
Thanks a lot for your pull request, looks good from a first quick view, but I'll make sure to give it a proper review and test later this week.

juliushaertl:
This is a really nice enhancement. Now from a UX perspective I think it would make sense to still somehow indicate that the rule is shown because it is inherited from a parent node. cc @jancborchardt 

image

The group2 entry is fully inherited from the parent (same as user1, but user1 was added as an individual rule to the "Child" directory). I would propose to make sure that the permission icons also get the same opacity level in that case.

nenadalm:
opacity fixed

jancborchardt:
> I would propose to make sure that the permission icons also get the same opacity level in that case.

Yep, good proposal. @nenadalm thanks for the contribution and the adjustment! :)

wiswedel:
so can this PR be approved and merged then, @juliushaertl?

juliushaertl:
Let me rebase so we can get this in

juliushaertl:
I've pushed a rebased version of this in https://github.com/nextcloud/groupfolders/pull/999 since i didn't have push access to your repo. Let me close this PR in favor of the new one then.

Issue #760, opened 2020-02-11
Denying permission for admin makes it impossible to recover folder


dipiana:
When you deny a group (where the admin is member of) access to a folder, admin can no longer see the folder / undo the changes - even though admin has advanced permissions.

Daksol:
Keep your admins as a separate group which only do advanced permissions settings. Danger else of locking yourself out.

fschrempf:
Closing this in favor of #1339.

Issue #759, opened 2020-02-11
Hide all subfolders expect one is explicitly shared


dipiana:
I am looking for a way to hide (deny access ) to all subfolders of the folder **projects**.
Only people who are granted access to some of these subfolders should be able to access them.

Therefore I tried to set advanced permission for the folder "projects" to deny and the permission of project_01 (which is inside "projects") to allow.

Unfortunately, the user does not have access to project_01.

Is there a way to only share specific projects within a folder?

Daksol:
See response to this item. It has arrangement of many subfolders at a level, but most users only see the one for their project (and would see two if part of two projects).

https://github.com/nextcloud/groupfolders/issues/1060

Issue #752, opened 2020-02-05
Allow stop inheritance of permissions


janpekar:
When I want to create subfolder inside GroupFolder I can only extend permissions. I don't see inherited permissions like mentioned in https://github.com/nextcloud/groupfolders/issues/724 and I cannot cancel this inheritance.
I can substitute this feature by creating GroupFolder inside existing GroupFolder. But it has disadvantages
- desktop sync is not working on subgroup folders - see https://github.com/nextcloud/desktop/issues/1773 and https://github.com/nextcloud/groupfolders/issues/652
- creating subGroupFolder sounds like workaround, GUI is not ready for that
- size of folder is not updated on parent folders, so quota is not applicable
- I need to administer it through NC settings and not easily through create folder and modify properties

My recommendation is to implement something what is known in NTFS world, where you can
- assign permission(s) only for current folder and don't inherit it to subfolders
- stop inheriting permissions from parent on particular folder and create permissions from the scratch on that folder (it is important to keep the way to assign permissions when no permissions are inherited)

Both of above options can solve many problems, only one can be implemented.

fschrempf:
I think what you want is possible without disabling inheritance. It's maybe not as straight-forward as it should be, but what about the workflow described here: https://github.com/nextcloud/groupfolders/issues/655#issuecomment-960823463. Does that solve your problem?

provokateurin:
All of the problems/drawbacks of having nested Groupfolders have been fixed, so I think we have a solution for this.

Issue #738, opened 2020-01-23
AP for all files in folder, but different AP for parent folder (eq. delete permission)


sim-jac:
We have the situation, that we share groupfolders with member-content, management-content and public-content of a group. To realize this we created three subfolder (member, management, public) and three groups with advanced permissions to the subfolder.

Furthermore wie deny the write/delete access for the three groups to the main-folder, to avoid, that user can create new folder beside the three subfolder we created. Of course user should also not be able to delete the subfolder. Unfortunately - if we set the advanced permission do deny deleting user are also not able do delete files _inside_ a subfolder.

Solution: The possibility of setting advanced-permission with wildcard to all files inside a folder, combined with different advanced-permissions of the parent-folder.

Best regards
Simon

Germar:
Related to #752 but different approach

fschrempf:
> Unfortunately - if we set the advanced permission do deny deleting user are also not able do delete files _inside_ a subfolder.

What's the issue with denying delete on the main directory and allow delete again in the three subdirectories?

PR #763, opened 2020-02-14
Extend permission assignment possibilities

nenadalm:
this pr includes https://github.com/nextcloud/groupfolders/pull/761

new featues:
- it is now possible to stop inheritance of advanced permission (there is "inherit" checkbox for that)
- it is possible to deny permissions to everyone by default, permission has to be granted using advanced permissions in that case (there is new "Default policy" select for this in settings)
- until now, there were admins that could manage permissions on files they had access to. There are new superadmins (currently it's only users in admin group - but multiselect can be added same way as for admins) that have read access to the whole group folder (recursively) so that if all permissions are denied by default, they can still manage them.

existing issues this will might solve:
- https://github.com/nextcloud/groupfolders/issues/752
- https://github.com/nextcloud/groupfolders/issues/738
- https://github.com/nextcloud/groupfolders/issues/760
- https://github.com/nextcloud/groupfolders/issues/759
- https://github.com/nextcloud/groupfolders/issues/724
  - in case "Default policy" is set to "Allow access", nobody can access folders
  - inherited advanced permissions are shown

please check if this pr solves Your problems. We are evaluating if it solves ours. In case functionality is ok, it might be wise to do some performance optimizations - might look into it later if it works as expected.


janpekar:
It solves problems for me. Creating nested groupfolders is no longer required.
I can create group folder and set Advanced permission for group, that has access which is not "inheritable" so subfolders hasn't that permission. For this to work is necessary "Allow access" feature because global group permissions set for group folder allow access for all members of that groups.

juliushaertl:
> it is possible to deny permissions to everyone by default, permission has to be granted using advanced permissions in that case (there is new "Default policy" select for this in settings)

I was wondering if we should do this on a per group level (similar to how the read/write/delete default permissions are set). We also should think about if we need to handle the root folder differently here, since otherwise you always need to add an ACL for the root itself first to make sure it shows up in the users home.

> until now, there were admins that could manage permissions on files they had access to. There are new superadmins (currently it's only users in admin group - but multiselect can be added same way as for admins) that have read access to the whole group folder (recursively) so that if all permissions are denied by default, they can still manage them.

I'm not entirely sure if the current code properly prevents users to adjust ACLs if they have limited access to a path due to other ACL limitations. Splitting users who can manage ACL and admin users adds a new level of complexity here, so for now I'd like to stick with the approach that there is no limitation for users who can manage ACLs and give those full read access to avoid unexpected behavior here.

nenadalm:
>> it is possible to deny permissions to everyone by default, permission has to be granted using advanced permissions in that case (there is new "Default policy" select for this in settings)

> I was wondering if we should do this on a per group level

We need to deny access by default and since it's just different way of doing it, I don't mind. Should I change that?

In case of using different policy for different groups I think my mind would blow up when trying to set permissions, but I have no intentions on doing so :)

> We also should think about if we need to handle the root folder differently here, since otherwise you always need to add an ACL for the root itself first to make sure it shows up in the users home.

Since it's consistent with other folders, I think it's fine.

Maybe if users have some permissions on sub folders/files then these users should have read access to root as well? If so, should this work also for other folders other than root? Say there is file `/root/f1/f2/file` and somebody adds read access to `file` to `user`? Should `user` have automatically read access on `f2`, `f1` and `root`? I would prefer if `root` wasn't special folder - less things to keep in mind.

>  I'd like to stick with the approach that there is no limitation for users who can manage ACLs and give those full read access to avoid unexpected behavior here.

ok. I'll change that.

cc @janpekar

jeltevdw:
What is the status of this pull request? It seems to be a very useful function.

Prometheorum:
Does this pull request still have a chance to be merged? This PR would allow much easier managment of group folders with high member count and complex permission structures.

fschrempf:
I don't think this PR has the potential to be merged in its current state. This doesn't mean that it doesn't add value. But rather than applying a pile of changes to cover multiple different issues, we probably want to work out the single points that need to be improved or fixed in the according issues and find solutions for each one separately.

We need to find solutions that don't add too much complexity and don't introduce too many (best: none) new user options that affect the overall behavior. Better fix the default behavior for all users instead.

One of the main points of this PR seem to be the issues mentioned in the description above. Let's discuss these topics there.

fschrempf:
I'm closing this due to the reasons mentioned above and invite everyone involved to discuss the solutions in the linked issues, create new issues if required and come up with PRs that solve one problem at a time.

Issue #808, opened 2020-03-19
GroupFolders API Create Folder request not Working


Benjamin-Forg:
Hi there,

I'm having a problem getting the API to work, I want to create a folder using this command from terminal:

`curl -X POST "https://username:password@domain-name/cloud/apps/groupfolders/folders" -d "mounpoint:testfolder" -H "OCS-APIRequest:true"`

But it just gives me an XML response listing all the folders. 

Also the DELETE requests give me a "405 Method Not Allowed" error, the only things I got to work were GET requests.

Am I doing something wrong?

Thanks for you help

RenWal:
It looks like nextcloud doesn't like the overloaded URLs (in the sense that there are multiple routes defined with different methods for the same URL). If you rename the API endpoint for addFolder in appinfo/routes.php from `/folders` to `/folders/add` then it works:

```
        [
                'name' => 'Folder#addFolder',
                'url' => '/folders/add',
                'verb' => 'POST'
        ],
```

RenWal:
For me the issue went away when specifying the path with `index.php`. In your example:
```
curl -X POST "https://username:password@domain-name/cloud/index.php/apps/groupfolders/folders" -d "mounpoint:testfolder" -H "OCS-APIRequest:true"
```

Benjamin-Forg:
Thanks for your help @RenWal, 

That fixed one problem but it still didn't work until I replaced the ``:`` with a ``=`` in ``"mountpoint:testfolder"`` so it works with:
<pre>curl -X POST "https://username:password@domain-name/cloud/index.php/apps/groupfolders/folders" -d "mountpoint=testfolder" -H "OCS-APIRequest:true"</pre>

PR #812, opened 2020-03-20
Expose Api also via OCS

to-kn:
refs #808

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -67,4 +67,71 @@ return ['routes' => [
 		'url' => '/folders/{id}/search',
 		'verb' => 'GET'
 	]
+], 'ocs' => [
+	[
+		'name' => 'Folder#getFolders',
+		'url' => '/folders',
+		'verb' => 'GET'
+	],
+	[
+		'name' => 'Folder#getFolder',
+		'url' => '/folders/{id}',
+		'verb' => 'GET'
+	],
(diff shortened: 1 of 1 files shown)


to-kn:
With these changes, the groupfolders API can be accessed via:

curl -X POST "https://username:password@domain-name/v2.php/apps/groupfolders/folders" -d "mounpoint:testfolder" -H "OCS-APIRequest:true"

appinfo/routes.php
```diff
@@ -52,6 +52,73 @@
 		'verb' => 'POST'
 	],
 
+	[
```
juliushaertl, line 55: There is no need to have those as regular routes, since Nextcloud can just reuse the OCS endpoints.

to-kn:
@juliushaertl you're welcome! Thank you for your comment - did i get it right, that you are suggesting to remove the 'routes' entry in favor of the 'ocs' entry?

juliushaertl:
Just the additional entries you added to the routes sections for now. We can also migrate the existing routes to ocs endpoints, but that requires adjustments in the frontend code that queries them, so I'd say this is something for a separate pull request.

juliushaertl:
I cherry-picked your commit to https://github.com/nextcloud/groupfolders/pull/997 and addressed my comments regarding an all-ocs-api. Thanks for getting this started.

PR #830, opened 2020-03-24
Fix: Trying to access array offset on value of type int 

kesselb:
at lib/private/Files/Node/Node.php#327. As reported at https://github.com/nextcloud/server/issues/19010 and similar to https://github.com/nextcloud/text/pull/608/commits/92df34121d057c95c246275b4fe46675eeaa3257. 

get() expects $path to be a string hence the file id is int or null.


juliushaertl:
No need to backport I'd say since 7.4 is 18 only.

PR #841, opened 2020-04-02
Fix bug

pingou2712:
It's fixes a bug... =) 
cacheEntry is not an array but an implemention of the interface OCP\Files\Cache\ICacheEntry


juliushaertl:
Thanks for your pull request. Could you tell us what kind of bug this fixes? Because the CacheEntry actually implements `\ArrayAccess` so the existing code should work just fine https://github.com/nextcloud/server/blob/5bf3d1bb384da56adbf205752be8f840aac3b0c5/lib/private/Files/Cache/CacheEntry.php#L30

pingou2712:
You're right, I send you a new PR

PR #844, opened 2020-04-03
Fix bug PatchICacheEntry

pingou2712:
Hi,
Fetch the correct cache entry if folder just created.

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -120,8 +120,8 @@ class MountProvider implements IMountProvider {
 
 	public function getMount($id, $mountPoint, $permissions, $quota, $cacheEntry = null, IStorageFactory $loader = null, bool $acl = false, IUser $user = null): IMountPoint {
 		if (!$cacheEntry) {
-			// trigger folder creation
-			$this->getFolder($id);
+			// trigger folder creation and get cacheEntry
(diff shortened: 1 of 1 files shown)


lib/Mount/MountProvider.php
```diff
@@ -120,8 +120,8 @@ private function getCurrentUID() {
 
 	public function getMount($id, $mountPoint, $permissions, $quota, $cacheEntry = null, IStorageFactory $loader = null, bool $acl = false, IUser $user = null): IMountPoint {
 		if (!$cacheEntry) {
-			// trigger folder creation
-			$this->getFolder($id);
+			// trigger folder creation and get cacheEntry
+			$cacheEntry = $this->getRootFolder()->getStorage()->getCache()->get($this->getFolder($id)->getId());
```
juliushaertl, line 124: getFolder might return null so this should be checked.

pingou2712:
> Do you also have some steps to reproduce the error?


It's been a long time that I don't touched this code... 
Sometimes through the web and, if my memory is good, always bug with command line via occ.

(sorry for my bad english, i'm french)

juliushaertl:
Thanks a lot for getting this started this. I've included your commit in https://github.com/nextcloud/groupfolders/pull/1346 and pushed some additional error handling to it so we can get this in.

PR #883, opened 2020-04-19
ACL Cache

Deltachaos:
We have enabled the ACL feature for hundreds of group folders in a large and heavy used installation. This had tripled the load on our database server because of a large number of additional database queries where required to generate the file lists.

If you have a lot of sync clients this gets even worse. The response times of the application are also increasing by a few hundreds of milliseconds.

Because of this i have implemented a cache for the ACL manager that keeps the actively used rules in the memory. Even when stored in Redis, this increases the performance dramatically.

### DIFF
--- a/lib/ACL/ACLRuleCache.php
+++ b/lib/ACL/ACLRuleCache.php
@@ -4,6 +4,7 @@ namespace OCA\GroupFolders\ACL;
 
 use OCA\GroupFolders\ACL\UserMapping\IUserMapping;
 use OCA\GroupFolders\ACL\UserMapping\IUserMappingManager;
+use OCA\GroupFolders\ACL\UserMapping\UserMapping;
 use OCP\ICache;
 use OCP\ICacheFactory;
 use OCP\IUser;
@@ -181,7 +182,9 @@ class ACLRuleCache
 		foreach ($rules as $rule) {
 			$rulePathKeys[] = $rulePathKey = $this->keyPathMapping($storageId, $path, $rule->getUserMapping());
 			$fileIdKeys[] = $ruleFileIdKey = $this->keyFileId($rule->getFileId());
-			$this->set($rulePathKey, $rule);
+			$this->set($rulePathKey, array_merge($rule->jsonSerialize(), [
+				'file_id' => $rule->getFileId()
+			]));
 		}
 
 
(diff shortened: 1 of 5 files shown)


Deltachaos:
I found a Bug that the cache is not cleared correctly sometimes that i still need to fix... So this should not be merged yet.

lib/ACL/ACLRuleCache.php
```diff
@@ -0,0 +1,259 @@
...
+	private $ruleCache;
+
+	/**
+	 * @var IUserMappingManager
+	 */
+	protected $userMappingManager;
+
+	public function __construct(ICacheFactory $cacheFactory, IUserMappingManager $userMappingManager) {
+		$this->userMappingManager = $userMappingManager;
+		$this->ruleCache = $cacheFactory->createDistributed('acl');
```
juliushaertl, line 32:
```suggestion
		$this->ruleCache = $cacheFactory->createDistributed('groupfolders-acl');
```

Maybe a bit safer in terms of naming collisions.

juliushaertl:
There might also be an issue with the cache invalidation if paths change e.g. through renaming with this approach, so it might be useful register some files hooks and invalidate then as well.

fschrempf:
Thanks @Deltachaos, this looks like a valuable contribution. Would you mind rebasing this to the current version, fix the remaining bugs and incorporate a solution for invalidating the cache on renaming events as proposed by @juliushaertl?

Deltachaos:
I am currently out of time to do this. Maybe this is something @JonathanTreffler from @verdigado can work on.

lib/ACL/ACLRuleCache.php
```diff
@@ -0,0 +1,259 @@
...
+	}
+
+	protected function set($key, $value, $ttl = null) : void
+	{
+		if ($ttl === null) {
+			$ttl = self::DEFAULT_TTL;
+		}
+
+		$ttl = rand($ttl, round($ttl * self::DERIVATION));
```
icewind1991, line 41: This looks wrong, this currently select a ttl between 0.1 days and 1 days, while I would expect a derivation of 0.1 to mean something like 0.9 days to 1 day

icewind1991 (changes requested):
Would also be great to have some explanation in a commit about how the data in the cache is stored/structured

fschrempf:
I also wonder if this should include some test coverage to make sure the caching and invalidation of the cache will continue to work properly in the future?

fschrempf:
@JonathanTreffler Do you or anyone else still care about getting this finished?

fschrempf:
Am I right to assume that #1694 already implements ACL caching as proposed here, at least part of it? If yes I will close this PR.

Issue #774, opened 2020-02-20
ACL: Access Control List missing in Nextcloud 18


rmeissn:
I run a Nextcloud instance and updated to v18 a few days ago. Since then I am missing the ACL view one the group folder. I checked that advanced permissions is ticked in settings for this folder and I remember granting such permissions on NC v16 or v17.

As you'll see in the screenshot there are no permissions set and no options to grant such - or am I missing something? There should be advanced permissions for this folder or at least the option to create such.

![Bildschirmfoto von 2020-02-20 13-21-15](https://user-images.githubusercontent.com/855967/74933388-206ec280-53e4-11ea-8fe5-43eac0aa89d3.png)

biva:
I have the same issue: any update?

biva:
For me, the problem was Firefox and not Nextcloud. I solved it by restarting Firefox with addons disabled (Menu, Help, ...). And then restarted Firefox with addons enabled. It works now. I can't say which addon was causing troubles :(

rmeissn:
@biva You're right! I tried your solution and the ACL menu is shown now.

tortuetorche:
Hi folks,

This workaround doesn't work for me with Firefox 74.0, with the last versions of Chromium and Chrome the menu is visible but only if the uBlock Origin extension is disabled for my Nextcloud website...

Have a good day,
Tortue Torche

rmeissn:
@tortuetorche @biva The day later the menu is gone again. At least for my instance uBlock (installed in Firefox and Chromium) isn't blocking anything. It works as expected using Chromium, but not using Firefox. Updating to Firefox 74 isn't improving anything.

langfingaz:
For me it helped to open a private Firefox browser window (with no previous cookies and no cached webcontent), then the advanced permissions were visible again.

caguiar:
I have the same problem.
I am configured as the user who has permissions to manage the advanced permissions on a group folder.
If I logout of Nextcloud and login again, and go straight to the folder, advanced permissions is visible. After being logged on for a while (don't know how long, but a few minutes) if I go back and the advanced permissions is no longer there.
So "workaround": Logout -> login.
Nextcloud: 18.0.4
Group Folders App: 6.0.6
Browser: Firefox 785 (on Ubuntu Linux)

mpfaff:
[...]

PR #917, opened 2020-05-19
Make files client extension more robust to loading race conditions

### LINKED ISSUE #894: ACL: Access Control List missing for group or multiple users
Nextcloud: 18.0.4
Group Folders App: 6.0.6
Browser: Firefox 75 (on Ubuntu Linux)
If I configure a single user as admin to set advanced permissions, this user is presented with the ACL list on the Share tab, and thus this user can configure advanced permissions.
If I configure a group, to which this same users belongs too, and set that group as admin to set advanced permissions, when this user goes to the Share tab, the ACL list is missing.
The same happens if instead of a sigle user, additional users are configured as admin to set advanced permissions, no ACL list.

juliushaertl:
This should fix #894 where sometimes during loading of the files list the `OCA.Files.App.fileList.filesClient` was not initialized when the extension of that happend to load the dav properties for ACLs.

Easier to review without whitespace changes: https://github.com/nextcloud/groupfolders/pull/917/files?diff=unified&w=1

### DIFF
--- a/src/client.js
+++ b/src/client.js
@@ -23,171 +23,179 @@
 import ACL_PROPERTIES from './model/Properties';
 import Rule from './model/Rule'
 
-_.extend(OC.Files.Client, ACL_PROPERTIES);
+let client
 
-// Allow nested properties in PROPPATCH
-// WIP branch at https://github.com/juliushaertl/davclient.js/tree/enhancement/nested-proppatch
-var patchClientForNestedPropPatch = function (client) {
-	client._client.getPropertyBody = function(key, propValue) {
-		var property = this.parseClarkNotation(key);
-		var propName;
+(() => {
 
-		if (this.xmlNamespaces[property.namespace]) {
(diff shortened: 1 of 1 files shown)


fri-sch:
Can you please merge this? #774, #894 are really annoying and render the ACL feature unusable.

By the way @juliushaertl: I'm not involved in development here, but instead of having cosmetic (whitespace) changes and functional changes/fixes in a single commit, you should separate them.
I guess it is quite common to consider this bad practice that makes it difficult to review and keep a clean history.

juliushaertl:
Unfortunately i have still encountered issues with the sidebar with this PR, so it needs some more work.

fri-sch:
> Unfortunately i have still encountered issues with the sidebar with this PR, so it needs some more work.

Ok, I hope this can be solved soon. If it would help I could probably do some testing as soon as a fix is available.

juliushaertl:
Ah nevermind, i branched off with the change included for stable18/stable19

PR #942, opened 2020-06-08
Make sure $path is a string

kesselb:
Similar to https://github.com/nextcloud/text/pull/485


lib/Mount/MountProvider.php
```diff
@@ -23,11 +23,9 @@
 
 use OC\Files\Storage\Wrapper\Jail;
 use OC\Files\Storage\Wrapper\PermissionsMask;
-use OCA\GroupFolders\ACL\ACLManager;
```
kesselb, line 26: Hmm. I think PHPStorm removed them on save. Are they required?

Issue #936, opened 2020-05-31
NotPermittedException


tobidwest:
The log files contain the following:

[core] Error: OCP\Files\NotPermittedException:  at <<closure>>

0. nextcloud/apps/groupfolders/lib/Versions/GroupVersionsExpireManager.php line 71
   OC\Files\Node\File->delete()
1. nextcloud/apps/groupfolders/lib/Versions/GroupVersionsExpireManager.php line 57
   OCA\GroupFolders\Versions\GroupVersionsExpireManager->expireFolder({id: 1,mount_poi ... e})
2. nextcloud/apps/groupfolders/lib/BackgroundJob/ExpireGroupVersions.php line 40
   OCA\GroupFolders\Versions\GroupVersionsExpireManager->expireAll()
3. nextcloud/lib/private/BackgroundJob/Job.php line 61
   OCA\GroupFolders\BackgroundJob\ExpireGroupVersions->run(null)
4. nextcloud/lib/private/BackgroundJob/TimedJob.php line 55
   OC\BackgroundJob\Job->execute(OC\BackgroundJob\JobList {}, OC\Log {})
5. nextcloud/cron.php line 125
   OC\BackgroundJob\TimedJob->execute(OC\BackgroundJob\JobList {}, OC\Log {})

tobidwest:
Does anyone know a work-around instead of disabling the app completely? 
The instance is useless at the moment, so any help would be greatly appreciated

Issue #861, opened 2020-04-09
Cannot delete files from Trashbin


lenaulrich:
When using groupfolders and deleting files they are stuck in the trash bin. When trying to delete a file from the trash bin permanently the message "Error while removing file from trashbin" is shown.

Error in the nextcloud.log:
{"Exception":"OCP\\Files\\NotPermittedException","Message":"","Code":0,"Trace":[{"file":"\/home\/user\/shared\/customer\/apps\/groupfolders\/lib\/Versions\/GroupVersionsExpireManager.php","l

Why this issue is caused:

- The file is marked as deleted and is listed in the database in the table: group_folders_trash with a link to the seemingly correct trash folder
- The file is not in the stated trash folder on the file system
- Therefore the file cannot be deleted

How to resolve this issue:

- Create the file in the correct trash bin folder
- Delete the file in the webinterface

Concerns Nextcloud version 17.0.3 (groupfolders version <= 5.05) and seems to be fixed in Nextcloud Version 17.0.5 (groupfolders version 5.0.5). After upgrading to Nextcloud version 17.0.5 the deleted files from Version 17.0.3 are still in the trashbin and cannot be deleted.

PHP Version 7.2

The issue cannot be solved by the occ command "files:scan" and "groupfolders:scan". Can you offer a solution for this issue?

horvan:
Can confirm the same issue for 18.3 php7.2 It seems to be either a mass deletion issue., Cant just delete one file at once. Had not reproduced it yet. Sorry.

Issue #673, opened 2019-11-10
NotPermittedException error


TP75:
FYI

[Groupfolders.app NotPermittedException error doesn’t stop](https://help.nextcloud.com/t/groupfolders-app-notpermittedexception-error-doesnt-stop/63426)

Needs triage and some investigation, I presume.

Issue #536, opened 2019-07-22
NotPermittedExecption Error


Bables55:
I started using Group folders actively with 4.0.4.
Then an error is regularly output to the log. This is probably Cron's timing.

I can only set the three permissions "Write, Delete, Share".  I think that I had the authority of "Create" before. Is this abolished?
I can not get the path when I execute Groupfolders:permission with OCC command.
Are these errors related to this error?

### Error log
```
[core] Error: OCP\Files\NotPermittedException:  at <<closure>>

0. /MY-NEXTCLOUD/apps/groupfolders/lib/Versions/GroupVersionsExpireManager.php line 71
   OC\Files\Node\File->delete()
1. /MY-NEXTCLOUD/apps/groupfolders/lib/Versions/GroupVersionsExpireManager.php line 57
   OCA\GroupFolders\Versions\GroupVersionsExpireManager->expireFolder({id: 6,mount_poi ... e})
2. /MY-NEXTCLOUD/apps/groupfolders/lib/BackgroundJob/ExpireGroupVersions.php line 40
   OCA\GroupFolders\Versions\GroupVersionsExpireManager->expireAll()
3. /MY-NEXTCLOUD/lib/private/BackgroundJob/Job.php line 61
   OCA\GroupFolders\BackgroundJob\ExpireGroupVersions->run(null)
4. /MY-NEXTCLOUD/lib/private/BackgroundJob/TimedJob.php line 55
   OC\BackgroundJob\Job->execute(OC\BackgroundJob\JobList {}, OC\Log {})
5. /MY-NEXTCLOUD/cron.php line 123
   OC\BackgroundJob\TimedJob->execute(OC\BackgroundJob\JobList {}, OC\Log {})
```

#### Server
- OS:
Linux 3.10.0-514.2.2.el7.centos.plus.x86_64 #1 SMP Wed Dec 7 19:10:15 UTC 2016 x86_64
- Webserver:
nginx/1.14.2 (fpm-fcgi)
- Database:
mysql 10.2.25
- PHP:
7.3.7
[...]

PR #990, opened 2020-08-07
Always delete expired versions regardless of the filecache permissions

juliushaertl:
If a version is created by a user which doesn't have delete permission the filecache entry for the version file has no delete permission set since the cache entry is copied over from the original file. https://github.com/nextcloud/groupfolders/blob/279eb30c27b16cfcfc89271da80a8d20b1664b9a/lib/Versions/VersionsBackend.php#L130

This PR makes sure that version expiry is always able to delete expired versions.

Fixes https://github.com/nextcloud/groupfolders/issues/861
Fixes https://github.com/nextcloud/groupfolders/issues/536
Fixes https://github.com/nextcloud/groupfolders/issues/673
Fixes https://github.com/nextcloud/groupfolders/issues/936

### DIFF
--- a/lib/Versions/GroupVersionsExpireManager.php
+++ b/lib/Versions/GroupVersionsExpireManager.php
@@ -23,6 +23,7 @@ namespace OCA\GroupFolders\Versions;
 
 
 use OC\Files\FileInfo;
+use OC\Files\View;
 use OC\Hooks\BasicEmitter;
 use OC\User\User;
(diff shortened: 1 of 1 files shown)


juliushaertl:
I think the idea approach would be to have a storage wrapper applying the correct permissions for the groupfolder versions as well, but for the sake of having a quick fix this PR is rather simple.

Issue #930, opened 2020-05-28
Take trashbin retention policy into account


juliushaertl:
Since the groupfolders app has its own trash backend, the regular background job for applying the retention policy does not act on files in the group folders trashbin: https://github.com/nextcloud/server/blob/afbd9c4e6ed834e713039f2cff88ba3eec03dadb/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php#L89

freiheitswolke-org:
Hey,
we opened a bounty on this missing function.
Please spread the voice and contribute as you can:
https://www.bountysource.com/issues/69919948-tashbin-cannot-be-cleared-with-group-folder-data

landryb:
is #1064 fixing that or just a temporary workaround to add to cron ?

juliushaertl:
#1064 is a temporary workaround to clear the whole trashbin, a proper retention implementation is started with #991

jngrb:
Please indicate what the development status is. Also in my project, nextcloud hosting for an NGO, we are in desperate need of a solution and would like how the Nextcloud team wants to proceed and whether we can help getting this fixed.

CarlSchwan:
New PR: https://github.com/nextcloud/groupfolders/pull/1759. It would be great to get some community testing.

fschrempf:
Closing in favor of #1759.

pgassmann:
FYI, group folders 11.1 for Nextcloud 23 is released. `occ groupfolders:expire` will now clean up the trashbin of groupfolders.

PR #991, opened 2020-08-07
Expire group folders trashbin

juliushaertl:
This is a first attempt to make group folder trash items expire properly, since until now the trash is just growing forever. 

Fixes https://github.com/nextcloud/groupfolders/issues/930

@icewind1991 In a later step, I'd like to extend `OCA\Files_Trashbin\Trash\ITrashBackend` with an expire method in the server, so that we can basically move the regular trashbin retention from the old static functions to the regular files trash backend as well. With this we can then pick up the retention policy globally and apply it to each trash backend. Any concerns regarding that?

Missing pieces for the groupfolder only implementation:
- [ ] Also take expiration based on available space into account
- [ ] Provide occ command and background job

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -22,12 +22,15 @@
 namespace OCA\GroupFolders\Trash;
 
 use OC\Files\Storage\Wrapper\Jail;
+use OC\User\User;
+use OCA\Files_Trashbin\Expiration;
 use OCA\Files_Trashbin\Trash\ITrashBackend;
 use OCA\Files_Trashbin\Trash\ITrashItem;
 use OCA\GroupFolders\ACL\ACLManagerFactory;
 use OCA\GroupFolders\Folder\FolderManager;
 use OCA\GroupFolders\Mount\GroupFolderStorage;
 use OCA\GroupFolders\Mount\MountProvider;
+use OCA\GroupFolders\Versions\VersionsBackend;
 use OCP\Constants;
 use OCP\Files\Folder;
 use OCP\Files\Node;
@@ -52,18 +55,23 @@ class TrashBackend implements ITrashBackend {
 	/** @var ACLManagerFactory */
 	private $aclManagerFactory;
 
+	/** @var VersionsBackend */
+	private $versionsBackend;
+
 	public function __construct(
 		FolderManager $folderManager,
 		TrashManager $trashManager,
 		Folder $appFolder,
 		MountProvider $mountProvider,
-		ACLManagerFactory $aclManagerFactory
+		ACLManagerFactory $aclManagerFactory,
+		VersionsBackend $versionsBackend
 	) {
 		$this->folderManager = $folderManager;
 		$this->trashManager = $trashManager;
 		$this->appFolder = $appFolder;
 		$this->mountProvider = $mountProvider;
 		$this->aclManagerFactory = $aclManagerFactory;
+		$this->versionsBackend = $versionsBackend;
 	}
 
 	public function listTrashRoot(IUser $user): array {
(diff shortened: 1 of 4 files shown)


icewind1991:
> so that we can basically move the regular trashbin retention from the old static functions to the regular files trash backend as well. With this we can then pick up the retention policy globally and apply it to each trash backend. Any concerns regarding that?

Sounds good, I didn't do it previously because I couldn't figure out a proper approach back then

lib/Trash/TrashBackend.php
```diff
@@ -313,4 +321,36 @@ public function getTrashNodeById(IUser $user, int $fileId) {
...
+					} catch (NotFoundException $e) {
+						$this->trashManager->removeItem($folderId, $groupTrashItem['name'], $groupTrashItem['deleted_time']);
+						continue;
+					}
+					if ($node->getStorage()->unlink($node->getInternalPath()) === false) {
+						throw new \Exception('Failed to remove item from trashbin');
+					}
+					$node->getStorage()->getCache()->remove($node->getInternalPath());
+					$this->trashManager->removeItem($folderId, $groupTrashItem['name'], $groupTrashItem['deleted_time']);
+					$this->versionsBackend->deleteAllVersionsForFile($folderId, $groupTrashItem['file_id']);
```
pelzvieh, line 348: This is going to cause an exception if file_id is NULL. In my installation this was true for about 25% of files, so it seems to be a relevant case.

```suggestion
					if (!is_null($groupTrashItem['file_id'])) {
						$this->versionsBackend->deleteAllVersionsForFile($folderId, $groupTrashItem['file_id']);
					}

```

juliushaertl: Ah right, the file_id was added at a later point https://github.com/nextcloud/groupfolders/blob/7d7c1247c59852f6e88f21ba1da07ad1c72a4c41/lib/Migration/Version501000Date20191218182434.php

Good catch.

neomilium:
Is there is any way to contribute to this feature?

Lars1963:
Is there any progress on this pr?

We use groupfolders quit heavily and the trashbin is increasely causing issues.

wimurk:
Is this still not fixed? my disk is 3 TB and we only use 500 GB. The trashbin uses 2.5TB !!!

jonasgarstick:
> Is this still not fixed? my disk is 3 TB and we only use 500 GB. The trashbin uses 2.5TB !!!

Unfortunately not, as far as I know. But you can use the following occ command to empty the trashbin manually:
`groupfolders:trashbin:cleanup`

wimurk:
@jonasgarstick 

> > Is this still not fixed? my disk is 3 TB and we only use 500 GB. The trashbin uses 2.5TB !!!
> 
> Unfortunately not, as far as I know. But you can use the following occ command to empty the trashbin manually:
> `groupfolders:trashbin:cleanup`

Ah thanks, i am running the command now. The `Delete all` doesn''t work but when you select single files, it does `work`.
I hope this will be fixed soon. At least a scheduler command to remove all files older then 7 days or something.

sbernhard:
What is the process here @juliushaertl ? 

Its not only a question about disk usage,  but also about GDPR in case private date is stored and not cleaned up. Of course, the cleanup occ command can be used but it will distinguish between files one day in trashbin or files which are older than a year.

ynott:
I find it very weird that there has been no progress against this PR.
Our customers (paid users with Nextcloud Standard support) are in trouble, but no action is being taken.
(I have already contacted the Nextcloud support portal.)

I understand that there are many things that need to be done in the development of Nextcloud.
However, if Nextcloud is an application for a file management system, I think it is a broken system to have endless accumulation of files in the trash.

We will hope you understand this matter.
@jospoortvliet

pelzvieh:
Indeed the Groupfolders plugin seems to be quite unfinished. However, there's a workaround for the missing integration with the Trash plugin:

1. ssh to your nextcloud installation server
2. run `sudo occ groupfolders:list` and take note of the "Folder ID"s for which you want to automatically purge the trash
3. create a file for daily cron using `sudo -e /etc/cron.daily/nextcloud-groupfolder-clean`
4. For each "Folder ID" <id> you want to purge, add a line reading `/usr/sbin/occ groupfolders:trashbin:cleanup -f <id>` (replacing <id> with the actual number you noted)
5. You might want to add `/usr/sbin/occ groupfolders:expire` as last line.

In result, the file might look like follows:
```
#!/bin/bash
/usr/sbin/occ groupfolders:trashbin:cleanup -f 9
/usr/sbin/occ groupfolders:trashbin:cleanup -f 13
/usr/sbin/occ groupfolders:expire
```

Of course, depending on the Linux distribution that your installation is based on, the actual paths might vary. Also it might be necessary to install cron support first (e.g.  if you're not using system cron as nextcloud cron mechanism).

ynott:
@pelzvieh 
That's a good work-around.

I think it is good to clean out the trash. Even in the real world.
Expiring my trash can in the real world is not a problem. 
Because I clean it up, and I know when I cleaned it up.

However, in the case of Nextcloud, there is a problem.
The user does not know when the trash was expired, and if the server deletes the trash at the end of each month, there is no way to undo it even if the user realizes at the beginning of the month that the deletion was a mistake.

I'm glad to see the NC22 being released.
I think it's great that new features are being added.

However, the newly added features will shine because Nextcloud is an excellent file manager, and file management is an important key feature of Nextcloud. Also, our customers mainly use the file management feature and pay for support on it.

Nextcloud's development resources are limited, and it is up to Nextcloud Gmbh to decide where to allocate them.

I would hope that some months out of the year you could spend some time fixing the file management (including the group folder feature) functionality.

fschrempf:
Superseded by #1759.

PR #1016, opened 2020-08-21
Improve performance querying group_folders_manage

### LINKED ISSUE #233: scaling issue when creating many groupfolders ?
For a specific usecase testing i'm creating 3k groupfolders via the API, assigning access to the same group. Ie more or less in a for loop in python code using requests module:

```
curl  -XPOST -H "OCS-APIREQUEST: true" -u admin https://xxx/apps/groupfolders/folders -dmountpoint=dirX
curl  -XPOST -H "OCS-APIREQUEST: true" -u admin https://xxx/apps/groupfolders/folders/$newgroupfolderid/groups -dgroup=groupA
```
what's interesting is that at the beginning of my loop (when, say, 500 folders were already created) the time spent on each request was around 0.9s, while after 1800 folders were created (and still running) the same requests take between 2.5 and 3s, after 2000 folders between 3 and 3.3s..

To me, it should be constant-time...

My guess is that *something* depends on the amount of existing groupfolders, maybe rescanning `$datadir/__groupfolders/` at each operation ? Will do the same test by creating 3k subfolders in a regular folder.. and have a look at the code to figure out where the bottleneck can be.

i dunno if that's to be expected, if that many folders is beyond what anyone tested, or if things can be improved, but that's a bit worrying.

jeffpardy:
We noticed a linear decrease in performance when querying group folders as more folders were created and tracked down the issue to querying `group_folders_manage` in the loop inside of `getAllFoldersWithSize`. This replaces the individual queries per folder with a single call to get all rows from `group_folders_manage`.

At 70k folders our average response times went from 4 minutes to 20 seconds after applying this change.

May close #233

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -110,9 +110,12 @@ class FolderManager {
 
 		$rows = $query->execute()->fetchAll();
 
+		$folderMappings = $this->getAllFolderMappings();
+
 		$folderMap = [];
 		foreach ($rows as $row) {
 			$id = (int)$row['folder_id'];
+			$mappings = $folderMappings[$id] ?? [];
 			$folderMap[$id] = [
 				'id' => $id,
 				'mount_point' => $row['mount_point'],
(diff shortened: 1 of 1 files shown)


lib/Folder/FolderManager.php
```diff
@@ -110,29 +110,48 @@ public function getAllFoldersWithSize($rootStorageId) {
 
 		$rows = $query->execute()->fetchAll();
 
+		$folderMappings = $this->getAllFolderMappings();
+
 		$folderMap = [];
 		foreach ($rows as $row) {
 			$id = (int)$row['folder_id'];
+			$mappings = $folderMappings[$id];
```
icewind1991, line 118:
```suggestion
			$mappings = $folderMappings[$id] ?? [];
```

to handle the case where there is no mapping for the folder

icewind1991 (changes requested):
seems good otherwise

jeffpardy:
Thanks @icewind1991, updated.

icewind1991 (approved):

juliushaertl (approved):

landryb:
Nice! Thanks

Issue #831, opened 2020-03-24
Deleting Files S3 Storage not working


JUVOJustin:
Hi, i have used groupfolders for a while now. If i recall correctly, i have issues **deleting files only in groupfolders since ~ NC18**.

I use **s3 as primary storage**. Deleting files in users home directory seems to work finde.
Here is the corresponding log:

``` [remote] Error: TypeError: Argument 1 passed to OC\Files\Cache\CacheQueryBuilder::whereFileId() must be of the type integer, null given, called in /var/www/clients/client1/web89/web/nextcloud/lib/private/Files/Cache/Cache.php on line 655 at <<closure>>

 0. /var/www/clients/client1/web89/web/nextcloud/lib/private/Files/Cache/Cache.php line 655
    OC\Files\Cache\CacheQueryBuilder->whereFileId(null)
 1. /var/www/clients/client1/web89/web/nextcloud/lib/private/Files/Cache/Wrapper/CacheWrapper.php line 199
    OC\Files\Cache\Cache->moveFromCache(OCA\TermsOfServi ... {}, "__groupfolders/1/_DSC0502.jpg", "__groupfolders/ ... 1")
 2. /var/www/clients/client1/web89/web/nextcloud/apps/groupfolders/lib/Trash/TrashBackend.php line 150
    OC\Files\Cache\Wrapper\CacheWrapper->moveFromCache(OCA\TermsOfServi ... {}, "/__groupfolders/1/_DSC0502.jpg", "__groupfolders/ ... 1")
 3. /var/www/clients/client1/web89/web/nextcloud/apps/files_trashbin/lib/Trash/TrashManager.php line 105
    OCA\GroupFolders\Trash\TrashBackend->moveToTrash(OCA\Files_Trashb ... l}, "_DSC0502.jpg")
 4. /var/www/clients/client1/web89/web/nextcloud/apps/files_trashbin/lib/Storage.php line 193
    OCA\Files_Trashbin\Trash\TrashManager->moveToTrash(OCA\Files_Trashb ... l}, "_DSC0502.jpg")
 5. /var/www/clients/client1/web89/web/nextcloud/apps/files_trashbin/lib/Storage.php line 100
    OCA\Files_Trashbin\Storage->doDelete("_DSC0502.jpg", "unlink")
 6. /var/www/clients/client1/web89/web/nextcloud/lib/private/Files/Storage/Wrapper/Wrapper.php line 266
[...]

PR #1048, opened 2020-09-16
Make sure to only move in cache if it was not already done by the storage

juliushaertl:
- S3 primary storage
- try to delete a file from a group folder

Before:
The deletion showed an error since the moveFromCache was already performed in https://github.com/nextcloud/server/blob/bb06b6cce46df881d84f34adfd5b1c3f315bd718/lib/private/Files/ObjectStore/ObjectStoreStorage.php#L362 and therefore the second move failed with the error from #831 

After:
The cache move attempt is only happening if the paths have not been updated already


juliushaertl:
@icewind1991 @rullzer :ping_pong: for review

rullzer (approved):
Looks sane

PR #1055, opened 2020-09-21
Use the proper paremter type for the IN query

rullzer:
Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>


juliushaertl (approved):
:see_no_evil:

PR #1073, opened 2020-10-08
Fix occ when files_trashbin is disabled

juliushaertl:
Otherwise loading the ITrashManager interface fails when injecting the TrashBackend. I'm open to more elegant suggestions on how to solve this, though there doesn't seem to be a way to register occ commands programmatically or I couldn't find it.

### DIFF
--- a/lib/Command/Trashbin/Cleanup.php
+++ b/lib/Command/Trashbin/Cleanup.php
@@ -46,10 +46,12 @@ class Cleanup extends Base {
 	 */
 	private $folderManager;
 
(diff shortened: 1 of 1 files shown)


juliushaertl:
@rullzer @icewind1991 :ping_pong: for review

enoch85:
Tested and works!

enoch85 (approved):

enoch85:
Without this fix, all updates to 20.0.1 are broken in the Nextcloud VM as it will break `occ` to run properly. 

I would really appreciate if this could get merged soon.

Thanks! :rocket:

rullzer:
ah joy. we should move this trashbin management to core I think at some point. So we can just depend it is there. That it is then disabled is fine.

PR #1155, opened 2020-11-23
Bug fix in user list selection

olidal:
The default limit in userSearch sets a maximum hard limit of 10 users (+ groups) in the drop-down list of known users when adding new GF permissions to a directory entry.
I changed the value to 100 which seems a more reasonable limit for larger instances..
Tested on NC 19 with GF 7.1.1.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -275,7 +275,7 @@ class FolderManager {
 		});
 	}
 
-	public function searchUsers($id, $search = '', $limit = 10, $offset = 0): array {
+	public function searchUsers($id, $search = '', $limit = 100, $offset = 0): array {
 		$groups = $this->getGroups($id);
 		$users = [[]];
 		foreach ($groups as $groupArray) {


olidal:
Sorry, this bug may not happen in all configs, I forgot to mention we use the LDAP backend, which comes itself with some hidden limits.

See here, for example, we have 18 registered users, but without setting the proper (higher) limit, the list gets truncated to 15 entries.
```
www-data@safer-nextcloud:~/nextcloud$ php occ ldap:search  '' | wc -l
15
www-data@safer-nextcloud:~/nextcloud$ php occ ldap:search --limit 20 '' | wc -l
18
```

And even though we set the value of `ldapPagingSize` to 500 has recommend in docs I found around (this parameter seems actually not doing much):

```
php occ ldap:show-config -- s01
+-------------------------------+--------------------------------------------------------------------------------------------+
| Configuration                 | s01                                                                                        |
+-------------------------------+--------------------------------------------------------------------------------------------+
<... 8< ... stuff snipped ... >8 ...>
| ldapPagingSize                | 500                                                                                        |
| ldapPort                      | 389                                                                                        |
<... 8< ... stuff snipped ... >8 ...>
```

juliushaertl:
Should be fixed by #1395

PR #1187, opened 2020-12-23
Move to IBootstrap

juliushaertl:


lib/Mount/GroupFolderStorage.php
```diff
@@ -65,7 +65,7 @@ public function getOwner($path) {
 	public function instanceOfStorage($class) {
 		// "implement" the interface without adding a hard dependency on nc15
 		if ($class === 'OCP\Files\Storage\IDisableEncryptionStorage') {
-			return true;
+			return false;
```
ChristophWurst, line 68: why?

juliushaertl: Woops not sure where that came from

juliushaertl: Fixed.

juliushaertl:
Tests will be fixed with https://github.com/nextcloud/groupfolders/pull/1188

ChristophWurst (approved):

rullzer (approved):

Issue #648, opened 2019-10-21
Collision detection for group folder target


rullzer:
With the revert in https://github.com/nextcloud/groupfolders/pull/647 we have the following issue again:

1. Create a folder 'foo' for user 'bar'
2. Add some files
3. Now an admin sets up a group folder 'foo'
4. 'bar' loses access to their original 'foo'

@icewind1991 any idea how to solve this?

PVince81:
also see https://github.com/nextcloud/groupfolders/issues/1246

ChristophWurst:
Potential fix at https://github.com/nextcloud/groupfolders/pull/1239

KodoGit:
Looks like #1246 leads here. The problem still exists with NC 21.0.4 docker (nextcloud:21-apache) with GF 9.0.3
Just updated from 20.

fschrempf:
> Looks like #1246 leads here. The problem still exists with NC 21.0.4 docker (nextcloud:21-apache) with GF 9.0.3

#1246 and #681 are about collisions between two groupfolders with the same name. This issue is still unresolved.
Whereas this issue seems to be about name collisions between a normal directory and a groupfolder. This issue is resolved.

dgsiegel:
> Whereas this issue seems to be about name collisions between a normal directory and a groupfolder. This issue is resolved.

@fschrempf unfortunately it does not look like this issue is resolved. I've managed to recreate @rullzer's scenario with subfolders:

- Create a team "foo"
- Create a team folder "foo" and give the team "foo" access to it
- Create a **normal** folder "bar" inside the "foo" group folder
- Create a new team folder "foo/bar"
- Now depending on who is having access to the "foo/bar" team folder, you'll either see the original normal folder "bar", or the team folder "bar", but you will not have access to both if you have access to both team folders.

This scenario would be relevant if you later on figure out that "bar" has some sensitive information in it and you want to restrict who is having access to it. Therefore you'd create a new team folder with a specific team/users.

fschrempf:
@dgsiegel Are you creating a nested groupfolder "foo/bar" inside the groupfolder "foo"? If yes, this might be a problem specific to the nesting of groupfolders.

Anyway, I stopped using Nextcloud altogether as I think many things are broken beyond repair. I'd recommend switching to OpenCloud if possible. They have "Spaces" built-in as a well-maintained core feature and not as an optional plugin (like NC groupfolders).

dgsiegel:
[...]

PR #1239, opened 2021-01-22
Check for naming conflicts before returning the user mounts

icewind1991:
New try for #523 by working on storage level instead of triggering the filesystem setup this should prevent issues caused by cyclic setup

Fixes https://github.com/nextcloud/groupfolders/issues/648


juliushaertl (approved):

juliushaertl:
Rebased and updated psalm baseline since the IMountPoint interface currently leaks a private class as return value:

```
ERROR: UndefinedDocblockClass - lib/Mount/MountProvider.php:107:20 - Docblock-defined class or interface OC\Files\Storage\Storage does not exist (see psalm.dev/200)
				$userStorage = $this->mountProviderCollection->getHomeMountForUser($user)->getStorage();
```

Issue #310, opened 2019-02-05
Is a slash/backslash in groupfolder name officially supported?


avychodil:
I want to have structure of shared folder in organisation so I have created groupfolers like:

company\folder name1
company\folder name2

that is work great and when user have folder company, the share will appear imediately. I just don't know whether it is supported function cause I didn't find it anywhere. To prevent future problems with it could anybody confirm it is an intention this behaviour and it will stay like this even in future versions.

Thank you,

Ales

avychodil:
I want to have structure of shared folder in organisation so I have created groupfolers like:

company\folder name1
company\folder name2

that is work great and when user have folder company, the share will appear imediately. I just don't know whether it is supported function cause I didn't find it anywhere. To prevent future problems with it could anybody confirm it is an intention this behaviour and it will stay like this even in future versions.

Thank you,

Ales

RogerWilco2015:
Hello avychodil,
I have tried the same thing.
It works both with slashes (/) and backslashed (\). If the user doesn't have a "company" folder (as per your example), you can add it as a GroupFolder.
Everything seems to work fine except the client app.
In fact in my experience changes that take place in that folder are not synchronized locally by the app.
Vice-versa works: a local change on the pc is propagated in the cloud.

If we could get this to work it would be great, the limitation of having all group folders attached to the root folder makes things very messy an cluttered when you have lots of them.

Can someone shed some light on this?
Thank you.

avychodil:
It's strange cause I've done it and it works great both on server and on local client (Mac OSX). The only problem is, whether it will be like this in future versions.

RogerWilco2015:
The problem arises only if I create the "company" folder as another Group Folder. If the "company" folder is created by the user then all works as expected in my case as well.

fschrempf:
We have at least two other related issues in #977 and #618 and I'm wondering myself if this is/should be supported.

@icewind1991 @juliushaertl Do you now if slashes/backslashes in folder names is officially supported? If not we should add code to prevent creating them and document this. If yes we need to care about potential bugs related to this.

icewind1991:
Note that there are 2 separate cases here, folder name with a `/` in to create subfolders such as in this issue, and the folder name being just a `/` causing the groupfolder to effectively replace the users home.

Both cases are considered supported but for the first case there is currently no support for ensuring the parent folder exist so admins should create both the parent folder and subfolder groupfolders separately.

[...]

Issue #1372, opened 2021-03-14
Document how to move groupfolders


fschrempf:
Apparently it is possible to move groupfolders by renaming them and using the path to the new location (e. g. see https://github.com/nextcloud/groupfolders/issues/1156#issuecomment-736563980).

For developers it is common that move and rename are the same operation, for others it is probably not and it would be helpful to have this in the documentation. A rather common usecase would be #1156.

danir-de:
This should definitely be treated as a feature.
This way it's possible to create more detailed structures with proper acl, that were previously only possible by using traditional file servers.

fschrempf:
It seems like nested groupfolders is officially supported, so we should have this special case documented in the README file.

PR #1380, opened 2021-03-19
add hint for subfolder groupfolders to readme

icewind1991:
Signed-off-by: Robin Appelman <robin@icewind.nl>

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,10 @@ For entitlements, only users from those groups are selectable which have to be c
 
 ## Notes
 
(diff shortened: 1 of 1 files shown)


fschrempf:
Closes #310

fschrempf (approved):
I think we need a bit more info about nested groupfolders for users to really understand this, but for now this is a good first step and the rest can be discussed in #1372.

juliushaertl (approved):

Issue #1096, opened 2020-10-27
Error when accessing ACLs of a group folder that has a non-existend group


ChristophWurst:
1) Create some groups
2) Create group folder with those groups
3) Delete one of the groups

Expected: code catches the missing one and proceeds

Actual: https://github.com/nextcloud/groupfolders/blob/72cc6b505c50e2af9069f2149ed49b2a31d05991/lib/Folder/FolderManager.php#L231-L235 throws an exception because a method is called on `null`.

ChristophWurst:
cc @juliushaertl

ChristophWurst:
Actually a duplicate of https://github.com/nextcloud/groupfolders/issues/633

jngrb:
I propose to fix this issue as implemented in my pull request #1395 . Besides this backend issue, my PR also fixes an issue interpreting the data in the frontend (src/components/SharingSidebarView.vue).

Issue #633, opened 2019-10-08
Error: Exception: Call to a member function getGID() on null


lesaff:
I am trying to add a group to a newly-created group folder. This is the error in the log.

Nextcloud version: `17.0.0.9`
Group folders version: `5.0.3`
PHP version: `7.3.10`

Error stack

```
[index] Error: Exception: Call to a member function getGID() on null at <<closure>>

0. /####REDACTED####/cloud/public/lib/private/AppFramework/App.php line 126
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\GroupFolders ... {}, "aclMappingSearch")
1. /####REDACTED####/cloud/public/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\GroupFolde ... r", "aclMappingSearch", OC\AppFramework\ ... {}, {id: "2",_route: ... "})
2. <<closure>>
   OC\AppFramework\Routing\RouteActionHandler->__invoke({id: "2",_route: ... "})
3. /####REDACTED####/cloud/public/lib/private/Route/Router.php line 297
   undefinedundefinedcall_user_func(OC\AppFramework\ ... {}, {id: "2",_route: ... "})
4. /####REDACTED####/cloud/public/lib/base.php line 1000
   OC\Route\Router->match("/apps/groupfolders/folders/2/search")
5. /####REDACTED####/cloud/public/index.php line 42
   OC::handleRequest()
```

andreasdeschka:
I have the same error. 

For me the reason was that I also use the following app:
Group SQL Backends https://apps.nextcloud.com/apps/user_sql

In our user external user database I changed the ID of a group, to which I added a rule in the group folders app.

To fix it, I had to go to the nexcloud database and set the ID of the group correctly in this table:
oc_group_folders_groups

The failure only appeared after I deleted the caches in Group SQL Backends.

Maybe t

ChristophWurst:
It happens when a group is removed. see more info in https://github.com/nextcloud/groupfolders/issues/1096

jngrb:
I propose to fix this issue as implemented in my pull request #1395 . Besides this backend issue, my PR also fixes an issue interpreting the data in the frontend (src/components/SharingSidebarView.vue).

jngrb:
@juliushaertl , this issue can be closed, I think. The PR is merged and the duplicate #1096 is also already closed.

PR #1395, opened 2021-03-27
Fixed searching for groups in the sharing sideview

jngrb:
The result of FolderManager::getGroups and ::searchGroups are indexed
arrays. Thus, the data returned by the .../search?... route
are JSON Objects and not arrays.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -255,13 +255,15 @@ class FolderManager {
 
 	private function getGroups($id): array {
 		$groups = $this->getAllApplicable()[$id] ?? [];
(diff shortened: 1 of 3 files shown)


jngrb:
I added a second commit that avoids calling getGID() on null when searching for available groups, referring to issues #633 and #1096 .

This other fix is in the backend and filters for any non-existing group reference before returning GID and displayname. This enables to continue editing the ACLs despite outdated group references in the groupfolder's permissions.

If desirable, I could also split the frontend and backend fixes into two PRs.

icewind1991 (approved):

juliushaertl (approved):

PR #1403, opened 2021-03-29
Revert "fix ACL error: user can not manage acl even assigned in setting"

juliushaertl:
Reverts nextcloud/groupfolders#1364 as it was merged without reviews and I don't see the need to fetch all rows just for checking existance the execute and rowCount should work fine. 

I've put branch protections for master/stable in place now as they seemed to be missing on the groupfolders repo.

@Nienzu Maybe you can reopen https://github.com/nextcloud/groupfolders/pull/1364 once the merge has been reverted and describe your setup a bit more in detail, especially what DB you are using, because I cannot reproduce the original issue with the code.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -275,7 +275,7 @@ class FolderManager {
 			->where($query->expr()->eq('folder_id', $query->createNamedParameter($folderId)))
 			->andWhere($query->expr()->eq('mapping_type', $query->createNamedParameter('user')))
 			->andWhere($query->expr()->eq('mapping_id', $query->createNamedParameter($userId)));
-		if (count($query->execute()->fetchAll()) === 1) {
+		if ($query->execute()->rowCount() === 1) {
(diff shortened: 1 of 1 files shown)


fschrempf:
I was lately looking at the recent changes in the repo and stumbled upon this change, too. The fact that it was merged without review and that all rows are fetched from the db just to count them also bothered me.

fschrempf (approved):

Nienzu:
I will describe the problem latter, sorry for the mistake.

fschrempf:
@juliushaertl Would you mind adding your "Signed-off-by" tag to the commit message so the [DCO check](https://github.com/nextcloud/groupfolders/pull/1403/checks?check_run_id=2216326172) will pass. Also I think it would be nice to have a short comment in the commit on why this was reverted just for later reference.

juliushaertl:
@Nienzu No problem. :) Feel free to propose a new PR so we can have another look then.

PR #1417, opened 2021-04-02
Fix/1413/allow admin delegation

### LINKED ISSUE #1413: Allow management delegation of groupfolders
Hi,

Company Arawa has identified a customer need where it is requested that non-NC admins are able to manage groupfolders (to simplify workflows and lower the load on IT staff).

After discussion with @jancborchardt and @juliushaertl it appears that NC would accept a change where access restrictions to groupfolders' REST API's would be managed in the same way as what's done in https://github.com/nextcloud/impersonate.

I'll send you a PR implementing this change in the next coming days.

Cc: @zak39 @fifh

# Alternate solution

During the discussion, @juliushaertl mentionned the possibility to rather create a stable PHP API  for groupfolders (which could then be used by any application regardless of the user's admin rights), but this solution has been dismissed for now and the implications of implementing it haven't been thought much.

StCyr:
This PR will implement admin delegation of groupfolders

Fixes #1413

TODO:

- [x] implement an admin setting to specify the groups allowed to manage groupfolders
- [ ] make groupfolders accessible for the specified groups
- [ ] make groupfolder's API accessible for the specified groups


StCyr:
err. I created this branch against stable21. closing and opening a new one

PR #1459, opened 2021-04-23
README.md: Be more blunt about the current status of the app

fschrempf:
We have to admit that are some major problems and hiding that from the users is not good.
Let's be more honest and let people know very clearly what they are up to.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Admin configured folders accessible by everyone in a group.
 
 ## Notes
 
-See [this pinned issue](https://github.com/nextcloud/groupfolders/issues/1414) to know the status of this app.
+:warning: **This app currently has some major bugs and drawbacks**: Before using this in production, make sure to read [this pinned issue](https://github.com/nextcloud/groupfolders/issues/1414) to get an idea of the current status. If you are able to help out in any way please get in touch.
 
 ## Configure folders
 


fschrempf:
@CarlSchwan As you reacted with thumbs-down, would you mind to give a short explanation for your objection? I will be happy to close this PR if someone can convince me it's a bad idea. I will probably close it anyway, due to a missing ack from someone at some point. :wink:

CarlSchwan:
> @CarlSchwan As you reacted with thumbs-down, would you mind to give a short explanation for your objection? I will be happy to close this PR if someone can convince me it's a bad idea. I will probably close it anyway, due to missing acks at some point. wink

I agree that the app is not bug-free but I have been working on fixing some big issues lately in it. Thanks a lot by the way for your help on doing bug triaging, it is really helpful.

And more importantly, the app is officially supported for Nextcloud enterprise and used by customers. So it is not something I would add to the description. of the app :)

fschrempf:
> I agree that the app is not bug-free but I have been working on fixing some big issues lately in it.

"Not bug-free" was a bit of an understatement in the past. But I agree that there is a positive trend. And seeing people like you fixing bugs is highly appreciated. Thanks!

> And more importantly, the app is officially supported for Nextcloud enterprise and used by customers. So it is not something I would add to the description. of the app :)

Yeah, and in my opinion it should never have been given that status. But it's not something that can be reverted now, so let's try to deal with the reality and get things fixed.

Issue #907, opened 2020-05-08
Trying to access array offset on value of type int at /var/www/nextcloud/lib/private/Files/Node/Node.php#327


chrissi55:
### Steps to reproduce
1. nc 18.0.4.2
2. have a partition like /boot/efi or maybe a system partition of type swap
3. check your logfiles  and the system monitoring of nc

### Expected behaviour
Tell us what should happen

No errors flooding the logs like

Error | PHP | Trying to access array offset on value of type int at /var/www/nextcloud/lib/private/Files/Node/Node.php#327
-- | -- | --

### Actual behaviour
Tell us what happens instead
Errors like above come periodically  every hour

### Server configuration

**Operating system**: ubuntu 18.04.4

**Web server:** NGinx 1.17.10

**Database:**PostgreSQL 12.2

**PHP version:**7.4.5

**Nextcloud version:** (see Nextcloud admin page) NC 18.0.4.2

**Where did you install Nextcloud from:**by WebUI Updater from 18.0.3

**List of activated apps:**

> Enabled:
>   - accessibility: 1.4.0
>   - activity: 2.11.0
>   - admin_audit: 1.8.0
>   - bruteforcesettings: 1.6.0
>   - calendar: 2.0.3
>   - cloud_federation_api: 1.1.0
>   - comments: 1.8.0
>   - contacts: 3.3.0
>   - dav: 1.14.0
>   - federatedfilesharing: 1.8.0
>   - federation: 1.8.0
>   - files: 1.13.1
>   - files_accesscontrol: 1.8.1
>   - files_automatedtagging: 1.8.2
>   - files_downloadactivity: 1.7.0
>   - files_external: 1.9.0
>   - files_pdfviewer: 1.7.0
>   - files_retention: 1.7.0
>   - files_rightclick: 0.15.2
>   - files_sharing: 1.10.1
>   - files_trashbin: 1.8.0
>   - files_versions: 1.11.0
>   - files_videoplayer: 1.7.0
>   - groupfolders: 6.0.6
>   - logreader: 2.3.0
>   - lookup_server_connector: 1.6.0
>   - nextcloud_announcements: 1.7.0
>   - notes: 3.3.0
>   - notifications: 2.6.0
>   - oauth2: 1.6.0
>   - onlyoffice: 4.1.4
>   - password_policy: 1.8.0
>   - photos: 1.0.0
>   - privacy: 1.2.0
>   - provisioning_api: 1.8.0
[...]

PR #1472, opened 2021-04-28
Enforce string for folder id when obtaining the trash folder

juliushaertl:
Fix for https://github.com/nextcloud/groupfolders/issues/907#issuecomment-821128558

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -212,7 +212,7 @@ class FolderManager {
 		$result->closeCursor();
 
 		return $row ? [
-			'id' => $id,
+			'id' => (int)$id,
 			'mount_point' => $row['mount_point'],
 			'groups' => isset($applicableMap[$id]) ? $applicableMap[$id] : [],
 			'quota' => $row['quota'],
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -230,7 +230,7 @@ class TrashBackend implements ITrashBackend {
 		$folders = $this->folderManager->getFoldersForUser($user);
 		foreach ($folders as $groupFolder) {
 			if ($groupFolder['folder_id'] === (int)$folderId) {
-				$trashRoot = $this->getTrashFolder($folderId);
+				$trashRoot = $this->getTrashFolder((int)$folderId);
 				try {
 					$node = $trashRoot->get($path);
 					if (!$this->userHasAccessToPath($user, $folderId . '/' . $trashItem->getOriginalLocation())) {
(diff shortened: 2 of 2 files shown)


acsfer (approved):
LGTM

kesselb:
https://github.com/nextcloud/groupfolders/blob/6633e64d198c2b042dbee8072ba8f148ddcc2855/lib/Trash/TrashBackend.php#L270-L272

To use strict types later we may cast the folder id to a string here. However i'm a bit supprised to see $folderId as integer here as the value comes from the database and should be a string.

fschrempf:
> However i'm a bit supprised to see $folderId as integer here as the value comes from the database and should be a string.

Looks like we already cast the id to int right after we get it from the db: 

https://github.com/nextcloud/groupfolders/blob/6e6d353754eab88c17c175942eef6d3380f3b73c/lib/Folder/FolderManager.php#L418

fschrempf (approved):
Casting the types back and forth (the folder id is originally a string when it is retrieved from the db) in several kind of places seems kind of messy, but cleaning this is out of scope of this PR.

kesselb:
> Casting the types back and forth (the folder id is originally a string when it is retrieved from the db) in several places seems kind of messy, but cleaning this is out of scope of this PR.

In groupfolders it's okay to treat the folderId as integer. As soon as we use the folderId to access the filesystem the folderId must be a string (because the filesystem api expects a string). 

A string typehint for getTrashFolder tell's php to convert the method parameter to a string. But only if strict types are not enabled for the current file. The goal is to have more files with strict code. Adding a cast to string before calling the typed method ensures that the code still works with strict mode enabled. 

https://github.com/nextcloud/groupfolders/blob/6633e64d198c2b042dbee8072ba8f148ddcc2855/lib/Trash/TrashBackend.php#L270

`$folderId = (string)$folder['folder_id'];` and it will also work with strict mode.

fschrempf:
> A string typehint for getTrashFolder tell's php to convert the method parameter to a string. But only if strict types are not enabled for the current file. The goal is to have more files with strict code. Adding a cast to string before calling the typed method ensures that the code still works with strict mode enabled.

Ok, sounds reasonable. But then it looks like there is at least one more case where `getTrashFolder()` is called with an int parameter. I guess this needs to be fixed, too:

https://github.com/nextcloud/groupfolders/blob/7d7abd6fa750d31d7afb8d284b58b513b6805b34/lib/Trash/TrashBackend.php#L323-L326

kesselb:
Yep. As getTrashFolder is a groupfolder method it would also work to typehint the method to int and cast the string inside getTrashFolder.

fschrempf:
@juliushaertl Could you improve this according to the suggestion from @kesselb? Thanks!

andyxheli:
Anyway we can get this on the next release ?

juliushaertl:
Adjusted accordingly to always use an int internally and only cast to string when referring to the nodes API.

I also added a cast to avoid possible further string occurences when building the folder array from the database results.

andyxheli (approved):

PR #1484, opened 2021-05-03
Only return user result once

juliushaertl:
- Have two groups with the same user being member of
- Search for the user in the ACL list

Before:
- See duplicate entries for the user

After:
- Each user result is only displayed once


StCyr (approved):
tested and works

PR #1534, opened 2021-06-16
Cancel ACL user/group search requests

juliushaertl:
Otherwise there might be issues where slow requests cause rerendering that does not match the search query


juliushaertl:
Failures unrelated #1537

skjnldsv (approved):

PR #1543, opened 2021-06-17
Mock database dependency as it is being pulled in by the CacheWrapper

### LINKED ISSUE #1537: CI failures due to invalid test
On some unrelated PR:
```

1) OCA\groupfolders\tests\ACL\ACLCacheWrapperTest::testHideNonRead
Your test case is not allowed to access the database.
```

@icewind1991

juliushaertl:
Test failed after https://github.com/nextcloud/server/pull/26874 was merged since the CacheWrapper now has a dependency on the database connection which is being pulled in through the server container.

Fixes #1537


PVince81 (approved):
:+1: assuming that the DB connection is not touched/relevant for this set of tests

otherwise need to add `@group DB` at the top of the file

juliushaertl:
Yep the CacheWrapperTest is just testing its internal code with mocks, apart from the constructor there does not seem to be a database dependency there.

PR #1560, opened 2021-06-18
Move to shared js build configs

juliushaertl:
Moved to the shared webpack config with slight additions to it for typescript/react support which is needed for the admin UI of groupfolders. Best to review in individual commits as the automated lint fixes were quite large. (af35b54 is only automated eslint fixing)


artonge (approved):
Looks good from an individual commit perspective. Would have been easier to review if split lint/updates

.github/workflows/node.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [12.x]
+        node-version: [14.x]
```
skjnldsv, line 16: Let's lose the matrix, so simplify the workflow names

juliushaertl: Adjusted

skjnldsv (approved):

PR #1588, opened 2021-07-08
ignore .idea/

ArtificialOwl:


juliushaertl:
That would rather be something for your local git excludes as it is specitic to the development workflow you're using:

https://git-scm.com/docs/gitignore#_description

> Patterns which are specific to a particular repository but which do not need to be shared with other related repositories (e.g., auxiliary files that live inside the repository but are specific to one user’s workflow) should go into the $GIT_DIR/info/exclude file.

fschrempf:
I'm closing this for the reasons given by @juliushaertl above.

PR #1597, opened 2021-07-17
PoC: create and use Circles instead of Groups

ArtificialOwl:
This is a quick and simple PoC that replace Groups with Circles.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -22,9 +22,25 @@
 namespace OCA\GroupFolders\Folder;
 
 use OC\Files\Cache\Cache;
-use OCA\Circles\CirclesManager;
+use OCA\Circles\Exceptions\CircleNotFoundException;
+use OCA\Circles\Exceptions\FederatedItemException;
+use OCA\Circles\Exceptions\FederatedUserException;
+use OCA\Circles\Exceptions\FederatedUserNotFoundException;
+use OCA\Circles\Exceptions\InitiatorNotFoundException;
+use OCA\Circles\Exceptions\InvalidIdException;
+use OCA\Circles\Exceptions\MemberNotFoundException;
+use OCA\Circles\Exceptions\OwnerNotFoundException;
+use OCA\Circles\Exceptions\RemoteInstanceException;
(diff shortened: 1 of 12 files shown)


ArtificialOwl:
The idea is to use SingleIds of the Circles App to add Group or Circle to the list of recipient.

Right now, only work from the `occ` command:

This is my list of available Circles:
Selection_047

We will use:

-  `aDh4c1kTgeQXnzl1MyKtp2uMRdJCjHI` as a normal Circle generated by a user,
- `nfVw7bGzGOyvYsnaotMwDqSYHUVNAiN` as a Nextcloud Group (the singleId is generated using `occ circles:sync`, or waiting for a background job)


### Adding Circles to the groupfolder:

>      $ ./occ groupfolder:group 2 nfVw7bGzGOyvYsnaotMwDqSYHUVNAiN
>      $ ./occ groupfolder:group 2 aDh4c1kTgeQXnzl1MyKtp2uMRdJCjHI

The singleId is stored in the database as it is, and used by the `FolderManager`

ArtificialOwl:
This is just a quick proof of concept, that will only be available in NC23. Lot of work to finalize this work:

- Implementation in core of an ocs API to allow a search/browse Circles from the front-end (ie. `GET /ocs/v1.php/cloud/circles/details`)
- Permissions,
- Migrations

lib/Folder/FolderManager.php
```diff
@@ -146,39 +172,64 @@ public function getAllFoldersWithSize($rootStorageId): array {
...
-		$query = $this->connection->getQueryBuilder();
-		$query->select('*')
-			->from('group_folders_manage');
+		$queryHelper = $this->circlesManager->getQueryHelper();
+		$query = $queryHelper->getQueryBuilder();
+
+		$query->select('folder_id', 'mapping_type', 'mapping_id')
+			  ->from('group_folders_manage', 'm');
+
+		$queryHelper->addCircleDetails('m', 'mapping_id');
```
juliushaertl, line 181: Depending on if circles is then mandatory or if groups can still be used (which I'd vote for), we should have a possible or move anything circles related to a dedicated services to also allow the app to work if circles is disabled.

PR #1633, opened 2021-08-17
 [Backport stable20]: Correctly calculate directory sizes when using an object store primary storage

CarlSchwan:
A object store storage should use an NoopScanner since it is not
possible to scan the directory structure of an object store. This makes
sure that the group folder mount point doesn't override the NoopScanner
that the object store storage use.


PVince81 (approved):
:+1: assuming you have retested it

juliushaertl (approved):

PR #1634, opened 2021-08-17
Fix fetching groups

CarlSchwan:
Before the api call was made to /ocs/v2.php/cloudgroups/details
and this is now changed to /ocs/v2.php/cloud/groups/details. This also
remove the compatibility with nextcloud <= 14

### DIFF
--- a/src/settings/Api.ts
+++ b/src/settings/Api.ts
@@ -45,19 +45,8 @@ export class Api {
 	}
 
 	listGroups(): Thenable<Group[]> {
-		const version = parseInt(OC.config.version, 10);
-		if (version >= 14) {
-			return $.getJSON(OC.linkToOCS('cloud', 1) + 'groups/details')
-				.then((data: OCSResult<{ groups: Group[]; }>) => data.ocs.data.groups);
-		} else {
-			return $.getJSON(OC.linkToOCS('cloud', 1) + 'groups')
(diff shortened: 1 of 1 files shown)


src/settings/Api.ts
```diff
@@ -46,17 +46,8 @@ export class Api {
...
-				.then((data: OCSResult<{ groups: Group[]; }>) => data.ocs.data.groups);
-		} else {
-			return $.getJSON(OC.linkToOCS('cloud', 1) + 'groups')
-				.then((data: OCSResult<{ groups: string[]; }>) => data.ocs.data.groups.map(group => {
-					return {
-						id: group,
-						displayname: group
-					};
-				}));
+		return $.getJSON(OC.linkToOCS('cloud', 1) + '/groups/details')
```
juliushaertl, line 49: I assume this failing due to the `@nextcloud/router` bump on server, right? So no backports needed.

CarlSchwan: I think it is. At least I wasn't encountering this error on stable21, let me try to check stable22 just to be sure.

CarlSchwan: Good that I checked it seems to fails on stable22 too, so will need to be backported to stable22. Do you maybe know why it now suddenly fails to build in the ci, locally it works?

PR #1654, opened 2021-09-03
Make advanced permissions to inherit per user/group

### LINKED ISSUE #1212: ACL inheritance not working as documented if user is member of multiple groups and one of them is read disabled
Hi,
we have nextcloud 20.0.4 with group folders 8.2.0 running for the local fire department. There is a problem if a user belongs to multiple groups and one of this group is read disabled for a folder.

It seems, if a user is **read denied from being a member of group via direct setting in ACL** for one folder, the user is still disabled even though being a member of an **another inherited allowed group** - this is wrong!

Believing the manual unset rights are always inherited from the parent folder and also allow overrules deny. But this is not working in our case!

We did a tryout scenario with reduced users and groups - you will find attached all informations as screenshots.

1. create test users and groups (OK -> see screenshot below)
2. create groupfolder (OK -> see screenshot below)
2. create ACL with occ (OK -> see screenshot below)
3. check in ACL in UI as admin (OK -> see screenshot below)
4. login as test users and check visibility of folders and subfolders (ERROR -> see screenshot below)

00_acl
09_acl_error

01_users
02_groupfolder

**Here you can see the ACL for the folders when logged in as admin - this is similar to the OCC settings:**

03_acl_test
04_acl_test_all
05_acl_test_department
06_acl_test_department_manager
07_acl_test_department_officer
08_acl_test_department_trainee

**Here you can see the UI when logged in as test_manager with the missing folder:**

10_error_test_manager

I hope to get feedback soon!

Greets Thorsten

YoitoFes:
I think that, in following case the member of `manager` and `test` should have write permission to `foo/bar/`.

| Folder path|Advanced permissions|
|---|:---|
| `foo/` | Allow all for `manager` group |
| `foo/bar/` | Deny write for `member` group |

But writing to `foo/bar` by the member is denied, because currently advanced permissions is calculated by the following steps:

1. Merge all user/group's permissions together with the rules that "allow" overwrites "deny", per depth of path
1. Merge permissions of all depth of path together with the rules that child's permissions overwrite parent folder's one.

Fix it by calculate permissions by following steps:

1. Merge permissions of all depth of path together with the rules that child's permissions overwrite parent folder's one, per user/group.
1. Merge all user/group's permissions together with the rules that "allow" overwrites "deny"

Close #1212

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Once configured, the folders will show up in the home folder for each user in th
 
 _Advanced Permissions_ allows entitled users to configure permissions inside groupfolders on a per file and folder basis.
 
-Permissions are configured by setting one or more of "Read", "Write", "Create", "Delete" or "Share" permissions to "allow" or "deny". Any permission not explicitly set will inherit the permissions from the parent folder. If multiple configured advanced permissions for a single file or folder apply for a single user (such as when a user belongs to multiple groups), the "allow" permission will overwrite any "deny" permission. Denied permissions configured for the group folder itself cannot be overwritten to "allow" permissions by the advanced permission rules.
+Permissions are configured by setting one or more of "Read", "Write", "Create", "Delete" or "Share" permissions to "allow" or "deny". Any permission not explicitly set for a user/group will inherit the user/group's permissions from the parent folder. If multiple configured advanced permissions for a single file or folder apply for a single user (such as when a user belongs to multiple groups), the "allow" permission will overwrite any "deny" permission. Denied permissions configured for the group folder itself cannot be overwritten to "allow" permissions by the advanced permission rules.
 
 ![advanced permissions](screenshots/acl.png)
 
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -122,10 +122,21 @@ class ACLManager {
 		$path = ltrim($path, '/');
 		$rules = $this->getRules($this->getParents($path));
 
-		return array_reduce($rules, function (int $permissions, array $rules) {
-			$mergedRule = Rule::mergeRules($rules);
-			return $mergedRule->applyPermissions($permissions);
-		}, Constants::PERMISSION_ALL);
+		$inheritedPermissionsByMapping = [];
+		array_walk_recursive($rules, function($rule) use(&$inheritedPermissionsByMapping) {
+			$mappingKey = $rule->getUserMapping()->getType() . '::' . $rule->getUserMapping()->getId();
+			if (!isset($inheritedPermissionsByMapping[$mappingKey])) {
+				$inheritedPermissionsByMapping[$mappingKey] = Constants::PERMISSION_ALL;
+			}
+			$inheritedPermissionsByMapping[$mappingKey] = $rule->applyPermissions($inheritedPermissionsByMapping[$mappingKey]);
+		});
+		if (empty($inheritedPermissionsByMapping)) {
+			return Constants::PERMISSION_ALL;
+		}
+
+		return array_reduce($inheritedPermissionsByMapping, function (int $mergedParmission, int $permissions) {
+			return $mergedParmission | $permissions;
+		}, 0);
 	}
 
 	/**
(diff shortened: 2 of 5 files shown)


thorsten-schwartz:
@YoitoFes Hi, seems to get not enough testing support. But we are still on the right track to hit the goal within one year after launching the original thread. Anyway, thanks for helping - you do all your best - free of charge. Thanks

CarlSchwan:
This might end up to be a problematic change. Groupfolders are often used is a way where all users have all access to the top folder and some subfolders are restricted for some groups.

With this change, the restricted folders would not be restricted anymore. This would be a big behavior change.

fschrempf:
> With this change, the restricted folders would not be restricted anymore. This would be a big behavior change.

Yes, the introduced behavior is more permissive than the current one. So there could be setups where this will lead to unwanted access for some users. We would need to be very clear about this breaking change if we release it.

Still, I think it is worth the effort as the current behavior is just not intuitive and has caused a lot of trouble in the past (see linked issues). Without this change, one needs to add rules for the "manager" group in each sub directory to explicitly grant them the permissions that in fact should be inherited from the parent. Applying this scheme on a directory tree with several (nested) sub directories and multiple "manager" and "user" groups leads to a ton of hard-to maintain ACL rules being applied. This in turn causes lots of confusion for anyone who maintains or needs to change ACL rules and increases the probability of errors in the ACL rules.

On the other hand, with this applied we would have an easy and coherent behavior. I would even treat this as a bugfix for the currently "broken" ACL rule handling.

Here is a extended view of the example from above to visualize the change in the most easy setup possible:

User | Groups
------ | ----------
userA | manager
userB | test
userC | manager, test

### Current behavior

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB | Resulting Permissions for userC
-- | -- | -- | -- | --
foo/ | Allow all for manager group | all | inherited from default permissions | all
foo/bar/ | Deny write for test group | all | -write | -write

### New behavior

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB | Resulting Permissions for userC
-- | -- | -- | -- | --
foo/ | Allow all for manager group | all | inherited from default permissions | all
foo/bar/ | Deny write for test group | all | -write | all

fschrempf:
Here is a extended view of the example from above to visualize the change in the most easy setup possible:

User | Groups
------ | ----------
userA | manager
userB | test
userC | manager, test

### Current behavior

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB | Resulting Permissions for userC
-- | -- | -- | -- | --
foo/ | Allow all for manager group | all | inherited from default permissions | all
foo/bar/ | Deny write for test group | all | -write | -write

### New behavior

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB | Resulting Permissions for userC
-- | -- | -- | -- | --
foo/ | Allow all for manager group | all | inherited from default permissions | all
foo/bar/ | Deny write for test group | all | -write | all

thorsten-schwartz:
@fschrempf @CarlSchwan 

Thanks for taking this issue. The basic topic is open for almost one year.
If you need further info don't hesitate asking me.

Greets Thorsten

icewind1991:
The problem I see here is that you can't deny access to a folder from a subset of the user that has allow permissions on the parent

User | Groups
------ | ----------
userA | manager
userB | manager

Current

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB
-- | -- | -- | -- 
foo/ | Allow all for manager group | all | all
foo/bar/ | Deny write for userB | all | -write

New

Folder path | Advanced permissions | Resulting Permissions for userA | Resulting Permissions for userB
-- | -- | -- | -- 
foo/ | Allow all for manager group | all | all
foo/bar/ | Deny write for userB | all | all

fschrempf:
> The problem I see here is that you can't deny access to a folder from a subset of the user that has allow permissions on the parent

That's indeed a problem. Maybe we can keep the behavior for ACL rules with a single user as subject and only adjust the behavior for group rules?

icewind1991:
That only partly resolves the issue. It all comes down to the intent of the admin when setting "conflicting" permissions with overlapping groups. In the original example, the admin could have intended to deny UserC from writing just as well as it could have been an accident.

I think "sub folder always overwrites parent permissions" is the best option as it still allows admins create both situations (in the original example, create a `+write` rule for `manager`  for `foo/bar`), while the proposed change only allows for one of the options (the one with more permissions).

thorsten-schwartz:
Hi developers,

are there any further actions in this topic? As I read in this thread you are discussing about several ideas to get rid of this problem. Please take a look on my basic problem once again whether  your suggestions would solve my problem.
As I wrote almost a year ago (one user is member of several groups which one is denied and one is allowed to the same folder).

I'm looking forward for your feedback - thanks!
Thorsten

rubentolosa:
I'm in Nextcloud 26 and Group folders 14.0.3 and I think I'm facing the same bug.

If I give advanced permissions (write) in a folder to a user that already is in a group without write permissions, he can write to that folder and "see" that has permissions inherited on subfolders. But it is not true, subfolders are not writtable for him.

vbier:
> I think "sub folder always overwrites parent permissions" is the best option as it still allows admins create both situations (in the original example, create a `+write` rule for `manager` for `foo/bar`), while the proposed change only allows for one of the options (the one with more permissions).

The question is do you want to have a solution that can handle 100% of the cases, but in reality is unusable? As soon as you have a tree stucture that is a little bit more complex, you will have to add +write rules to hundreds of folders. I would even go so far to say, that this makes permission inheritance abundant. As you do not know which groups might overlap (and that can change any time by someone adding new users to a group), you will then have to effectively add a complete set of permissions for every group to every folder. 

We planned to use nextcloud to host our 850000 files in 260000 folders, with around 300 groups. We have 4600 permission assignments if the proposed change would be implemented. If it stays as it is now, I would have to assign the permissions for every folder and group, which would mean 78 million permission definitions. Aside from the fact that this will probably affect performance, this would be completely unmanageable for admins.

If it would be changed, it would be easy to use and understand. The fact that there are bug reports again and again indicates that the current imlementation seems to be unexpected to a lot of users. It sure does not seem logical to me that a user that is added to another group has less permissions than before.

Issue #694, opened 2019-12-12
File Versions do not work properly in groupfolders


hriekehof:
Hi there,

at the moment i cannot get file versions to work properly for files in groupfolders. In normal none shared personal folders it seems to work.

I use an excel sheet for testing.
1. Put an excel sheet into a groupfolder
2. Sync it via desktop client.
3. Save a second version and sync it via desktop client
4. Restore the previous version in Nextclouds web interface
5. Open excel sheet on desktop synced via desktop app. Its not changed to the previous version

Current Installation:
- Nextcloud 16.0.6 as docker container apache
- Groupfolder App 4.1.4
- Used Browser: Latest Chrome
- Desktop Client: 2.6.1
- Windows 10 Pro latest updates

I really would like if versions would start to work in groupfolders since we use them alot in our installation. 

What is also strange that whenever i hit the versions button a new version seems to be created in the list of versions
![image](https://user-images.githubusercontent.com/9348635/70708201-42f4a900-1cda-11ea-8871-532a1121fc71.png)

I would expect when restoring a previous versions that the list just gets reordered.

wiswedel:
@hriekehof what desktop app version are you using?

wiswedel:
confirmed on NC 17.0.1 with client 2.6.1 (Linux)

hriekehof:
hi i updated the comment i saw i had a mistake there :) I use Nextcloud on Win 10 Pro and i have the desktop client 2.6.1

wiswedel:
CC @nextcloud/server-triage

nickvergessen:
cc @icewind1991 

Although I think I remember versions being a topic for 18 only?

alexanderdd:
Does anybody know if versions work properly with group folders?

I need to migrate all of our folders to group folders, but if versions don't work, I won't migrate.

acsfer:
@alexanderdd it seems it does'nt as per https://github.com/nextcloud/groupfolders/issues/694

CarlSchwan:
#1660 should have fixed some issues with rolling back versions. This has been backported to 20, 21 and 22. I'm gonna close this issue but feel free to reopen if the problem persists.

PR #1660, opened 2021-09-09
Fix rollback of file version

artonge:
## Error

After the rollback, the `filecache` entry for the file does not contain the correct size which breaks the download.

## Current behaviour when a file `A` is rolled-back from `A.version1`

1. A new version is created
    - file `A` is copied to version `A.version2`
    - file cache `A` is copied to version cache `A.version2`

2. Version1 is restored
    - version `A.version1` is copied to file `A`
    - file cache `A.version1` is copied to version cache `A.version1` (which does nothing)

## Proposed change

The PR changes the last step to: file cache `A.version1` is copied to version cache `A`

### DIFF
--- a/lib/Versions/VersionsBackend.php
+++ b/lib/Versions/VersionsBackend.php
@@ -123,7 +123,7 @@ class VersionsBackend implements IVersionBackend {
 			$versionInternalPath = $version->getVersionFile()->getInternalPath();
 
 			$targetMount->getStorage()->copyFromStorage($versionMount->getStorage(), $versionInternalPath, $targetInternalPath);
-			$versionMount->getStorage()->getCache()->copyFromCache($targetCache, $versionCache->get($versionInternalPath), $versionInternalPath);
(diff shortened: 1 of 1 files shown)


fschrempf:
I assume this is related to #694. Just for reference.

artonge:
> I assume this is related to #694. Just for reference.

Nice, I did not look for related issues. Thanks for linking :)

CarlSchwan:
With the latest revision of this MR, the cache should now be copied to the correct location :D

artonge:
Tested, and it works :)

PVince81 (approved):
:+1: makes sense.

can you rebase to make the branch up to date so we can merge ?

PVince81:
stable22: https://github.com/nextcloud/groupfolders/pull/1675

PVince81:
stable21: https://github.com/nextcloud/groupfolders/pull/1674

Issue #1693, opened 2021-10-15
Slow performance with ACL enabled and many entries


PVince81:
### Steps

1. Create a folder "local/sub"
2. Setup a group folder "groupfolders/sub"
3. Setup another group folder "groupfolders_acl/sub" and enable ACLs
4. In all three "sub" folders, create 1100 folders (trick: capture a MKCOL and copy as curl, then put it in a loop on the CLI)
5. Open a folder in web UI and copy the PROPFIND from the network console as curl (to have the PROPFIND body ready)
6.. Adjust the curl command to do the PROPFIND on "local/sub" and run it with `time`
7. Adjust the curl command to do the PROPFIND on "groupfolders/sub" and run it `time`
8. Adjust the curl command to do the PROPFIND on "groupfolders_acl/sub" and run it `time`

### Expected result
Acceptable performance for all three calls.

### Actual result
The calls to "local/sub" and "groupfolders/sub" took about 1.5s
The call to "groupfolders_acl/sub" took 15 seconds!

### Versions
Nextcloud 22.1.1
groupfolders 10.0.0

### Analysis

Possible related to the code that resolves the ACLs

PR #1694, opened 2021-10-15
Use already cached parents when fetching ACL

PVince81:
When fetching ACL rules, don't refetch the rules from already cached
parents. This fixes performance issues with large folders.

Fixes https://github.com/nextcloud/groupfolders/issues/1693

Improves performance with 1100 entries from 20 seconds to 4 seconds.

- [x] BUG: because the cache is capped, the logic breaks after cached entries are getting removed by the cap (when testing I increased the capped cache's max)

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -54,40 +54,31 @@ class ACLManager {
 		return $this->rootStorageId;
 	}
 
-	private function pathsAreCached(array $paths): bool {
-		foreach ($paths as $path) {
-			if (!$this->ruleCache->hasKey($path)) {
-				return false;
-			}
-		}
-		return true;
-	}
-
 	/**
 	 * @param int $folderId
 	 * @param array $paths
 	 * @return (Rule[])[]
 	 */
 	private function getRules(array $paths): array {
-		if ($this->pathsAreCached($paths)) {
-			$rules = array_combine($paths, array_map(function (string $path) {
-				return $this->ruleCache->get($path);
-			}, $paths));
-		} else {
-			$rules = $this->ruleManager->getRulesForFilesByPath($this->user, $this->getRootStorageId(), $paths);
-			foreach ($rules as $path => $rulesForPath) {
-				$this->ruleCache->set($path, $rulesForPath);
-			}
+		// beware: adding new rules to the cache besides the cap
+		// might discard former cached entries, so we can't assume they'll stay
+		// cached, so we read everything out initially to be able to return it
+		$rules = array_combine($paths, array_map(function (string $path) {
+			return $this->ruleCache->get($path);
+		}, $paths));
 
-			if (count($paths) > 2) {
-				// also cache the direct sibling since it's likely that we'll be needing those later
-				$directParent = $paths[1];
-				$siblingRules = $this->ruleManager->getRulesForFilesByParent($this->user, $this->getRootStorageId(), $directParent);
-				foreach ($siblingRules as $path => $rulesForPath) {
-					$this->ruleCache->set($path, $rulesForPath);
-				}
+		$nonCachedPaths = array_filter($paths, function (string $path) use ($rules) {
+			return !isset($rules[$path]);
+		});
+
+		if (!empty($nonCachedPaths)) {
+			$newRules = $this->ruleManager->getRulesForFilesByPath($this->user, $this->getRootStorageId(), $nonCachedPaths);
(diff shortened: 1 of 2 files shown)


lib/ACL/ACLManager.php
```diff
@@ -54,40 +54,48 @@ private function getRootStorageId() {
...
+		// beware: adding new rules to the cache besides the cap
+		// might discard former cached entries, so we can't assume they'll stay
+		// cached, so we read everything out initially to be able to return it
+		$rules = array_combine($paths, array_map(function (string $path) {
+			return $this->ruleCache->get($path);
+		}, $paths));
+
+		$nonCachedPaths = [];
+		foreach ($paths as $path) {
+			if (!$this->ruleCache->hasKey($path)) {
```
come-nc, line 72: This seems redundant with the array_combine above, should be `$rules[$path] !== null` or `isset($rules[$path])` no?
`$nonCachedPaths` may even be computed by array_filter but not sure about the performance impact.

lib/ACL/ACLManager.php
```diff
@@ -54,40 +54,48 @@ private function getRootStorageId() {
...
 			}
+		}
 
-			if (count($paths) > 2) {
-				// also cache the direct sibling since it's likely that we'll be needing those later
-				$directParent = $paths[1];
+		if (count($paths) > 2) {
+			// also cache the direct sibling since it's likely that we'll be needing those later
+			$directParent = $paths[1];
+			if (!$this->ruleCache->hasKey($directParent)) {
```
come-nc, line 90: This is only possible if we just hit the cap right?

lib/ACL/ACLManager.php
```diff
@@ -54,40 +54,48 @@ private function getRootStorageId() {
...
+		}, $paths));
+
+		$nonCachedPaths = [];
+		foreach ($paths as $path) {
+			if (!$this->ruleCache->hasKey($path)) {
+				$nonCachedPaths[] = $path;
+			}
+		}
+
+		\OC::$server->getLogger()->debug('## getRules: "' . json_encode($paths). '" noncached=' . json_encode($nonCachedPaths), ['app' => 'core']);
```
PVince81, line 77: note to self: remove this

CarlSchwan:
Unit tests are not happy and could you rebase?

lib/ACL/ACLManager.php
```diff
@@ -54,40 +54,43 @@ private function getRootStorageId() {
...
 			}
+		}
 
-			if (count($paths) > 2) {
-				// also cache the direct sibling since it's likely that we'll be needing those later
-				$directParent = $paths[1];
+		if (count($paths) > 2) {
+			// also cache the direct sibling since it's likely that we'll be needing those later
+			$directParent = $paths[1];
+			if (!isset($rules[$directParent])) {
```
come-nc, line 85: I still do not understand, if `ruleCache->get($paths[1])` is null, `$paths[1]` is put in `$nonCachedPaths`, and `ruleManager->getRulesForFilesByPath` is called on it on line 75, and the result is stored in `$rules`.
Is it still possible for `$rules[$paths[1]]` to not be set after that?

PVince81: I've just investigated that with a debugger, and after calling `getRulesFoFilesByPath` I saw that all the keys will be populated. If no rules, there will be an empty array: https://github.com/nextcloud/groupfolders/blob/bug/1693/fix-acl-perfs/lib/ACL/RuleManager.php#L116

but I get your point, I also wonder why we'd need that second block at all. I'll recheck with the original code

PVince81: alright, got my mistake. the block isn't loading a rule for `$directPath` but it's doing so for the siblings.
the call is on `getRulesForFilesByParent` not `getRulesForFilesByPath`.

since I don't know the sibling names I can't do a cache check, so I'll simply let it through and hope the perf issue doesn't appear here after testing

PVince81:
back to draft mode

thanks to @come-nc for pointing out my thinking mistake

I had disabled siblings pre-fetching which is likely why everything got faster.
I've reenabled siblings pre-fetching and now it's up to 12 seconds again.

12 seconds is still better than 20s, so it means my original optimization did help, but it's still too much.

I'll try and understand siblings prefetching better and see if it fits this use case.
If it does fetch all the 1200 entries in one go, it would be nice because then everything gets cached in one go.
But it doesn't seem so, otherwise performance would be even better.

Needs further research...

PR #1753, opened 2021-11-14
Enfore the usage of int for quota

CarlSchwan:
The Doctrine DBAL returns a string for a bigint. This causes problems
since most of the code expects an int (QuotaStorage and MountProvider::getMount).

In particular the BackgroundJob for version expiration got broken
because of that.

The only disadvantage is that it will theorically breaks on 32bits
system but since we don't support that and some part of the code were
already expecing an int. This probably won't change anything.

@PhMetzger


juliushaertl (approved):

acsfer:
@CarlSchwan any reason to don't backport this to 22/21?

PR #1793, opened 2021-12-02
Cast argument from string to int

artonge:
The function expects an int but is given a string.

Is this a safe way to cast a string in php?

### DIFF
--- a/lib/Command/Scan.php
+++ b/lib/Command/Scan.php
@@ -55,7 +55,7 @@ class Scan extends Base {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
-		$folderId = $input->getArgument('folder_id');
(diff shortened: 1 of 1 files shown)


lib/Command/Scan.php
```diff
@@ -56,7 +56,7 @@ protected function configure() {
 
 	protected function execute(InputInterface $input, OutputInterface $output) {
 		$folderId = $input->getArgument('folder_id');
```
nickvergessen, line 58:
```suggestion
		$folderId = (int) $input->getArgument('folder_id');
```

cast here, so that the variable is save to use for potential future usage?

artonge: Make sense

nickvergessen (approved):

CarlSchwan (approved):

Issue #1702, opened 2021-10-22
NC22 Doctrine\DBAL\Query\QueryException: More than 1000 expressions in a list are not allowed on Oracle


toriva:
Good day! I hope I posted the message correctly.
Please help. After NC auto-update via snap to version 22, an error appeared and synchronization works intermittently. Please tell me what to do? Thank you in advance.

### Server configuration

**Operating system:** Ubuntu 20.04.2 LTS x86_64

**Web server:** Apache (fpm-fcgi)

**Database:** mysql 8.0.26

**PHP version:** 8.0.11

Modules loaded: Core, date, libxml, openssl, pcre, zlib, bcmath, bz2, ctype, curl, dom, hash, fileinfo, filter, ftp, gd, gmp, SPL, iconv, intl, json, ldap, mbstring, pcntl, PDO, session, posix, Reflection, standard, SimpleXML, mysqlnd, exif, tokenizer, xml, xmlreader, xmlwriter, zip, pdo_mysql, cgi-fcgi, redis, Zend OPcache

**Nextcloud version:** 22.2.0 - 22.2.0.2

**Group folders version:** 10.0.0

**Updated from an older Nextcloud/ownCloud or fresh install:** auto-updated via snap

**Are you using encryption:** no

### Logs

```
Doctrine\DBAL\Query\QueryException: More than 1000 expressions in a list are not allowed on Oracle.
/var/snap/nextcloud/28586/nextcloud/extra-apps/groupfolders/lib/ACL/RuleManager.php - line 113:
OC\DB\QueryBuilder\QueryBuilder->execute()
/var/snap/nextcloud/28586/nextcloud/extra-apps/groupfolders/lib/ACL/ACLManager.php - line 77:
OCA\GroupFolders\ACL\RuleManager->getRulesForFilesByPath()
/var/snap/nextcloud/28586/nextcloud/extra-apps/groupfolders/lib/ACL/ACLManager.php - line 118:
OCA\GroupFolders\ACL\ACLManager->getRules()
/var/snap/nextcloud/28586/nextcloud/extra-apps/groupfolders/lib/ACL/ACLCacheWrapper.php - line 98:
OCA\GroupFolders\ACL\ACLManager->preloadPaths()
/var/snap/nextcloud/28586/nextcloud/extra-apps/groupfolders/lib/ACL/ACLCacheWrapper.php - line 68:
OCA\GroupFolders\ACL\ACLCacheWrapper->preloadEntries()
[...]

PR #1804, opened 2021-12-07
fix Oracle query limit compliance

to-kn:
these changes fixed the Problems with #1702 for me

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -453,10 +453,16 @@ class FolderManager {
 				'a',
 				$query->expr()->eq('f.folder_id', 'a.folder_id')
 			)
-			->where($query->expr()->in('a.group_id', $query->createNamedParameter($groupIds, IQueryBuilder::PARAM_STR_ARRAY)));
+			->where($query->expr()->in('a.group_id', $query->createParameter('groupIds')));
 		$this->joinQueryWithFileCache($query, $rootStorageId);
 
-		$result = $query->executeQuery()->fetchAll();
+		// add chunking because Oracle can't deal with more than 1000 values in an expression list for in queries.
+		$result = [];
+		foreach (array_chunk($groupIds, 1000) as $chunk) {
+			$query->setParameter('groupIds', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
+			$result = array_merge($result, $query->executeQuery()->fetchAll());
+		}
+
 		return array_map(function ($folder) {
 			return [
 				'folder_id' => (int)$folder['folder_id'],


fschrempf (changes requested):
Thanks for the contribution! I think it would be better to add the chunking directly in `getFoldersForGroups()` instead of implementing it in multiple places. And a comment with the reason for the chunking would be nice.

fschrempf:
You can have a look at https://github.com/nextcloud/server/pull/27187 or https://github.com/nextcloud/mail/pull/4947 for other places where this was fixed. I also opened an issue (#1805) to check any other queries that might hit the limitations.

to-kn:
HI @fschrempf do you mean like that? To be honest i did not had the possibility to validate the fix like that...

fschrempf (changes requested):
The code looks good to me. If you could add the comment I was asking for and squash the two commits into a single one, I think we are good to go.

fschrempf:
And we nee to find out why the unit tests fail. Might not be related to this PR, I haven't checked. Maybe rebasing already does the trick.

to-kn:
@fschrempf should be good

lib/Folder/FolderManager.php
```diff
@@ -631,7 +637,7 @@ public function setFolderACL(int $folderId, bool $acl): void {
 	 */
 	public function getFoldersForUser(IUser $user, int $rootStorageId = 0): array {
 		$groups = $this->groupManager->getUserGroupIds($user);
-		$folders = $this->getFoldersForGroups($groups, $rootStorageId);
+        $folders = $this->getFoldersForGroups($groups, $rootStorageId);
```
CarlSchwan, line 640:
```suggestion
       $folders = $this->getFoldersForGroups($groups, $rootStorageId);
```

lib/Folder/FolderManager.php
```diff
@@ -654,7 +660,7 @@ public function getFoldersForUser(IUser $user, int $rootStorageId = 0): array {
 	 */
 	public function getFolderPermissionsForUser(IUser $user, int $folderId): int {
 		$groups = $this->groupManager->getUserGroupIds($user);
-		$folders = $this->getFoldersForGroups($groups);
+        $folders = $this->getFoldersForGroups($groups);
```
CarlSchwan, line 663:
```suggestion
       $folders = $this->getFoldersForGroups($groups);
```

CarlSchwan (approved):
Code looks fine, let's wait for the tests results

CarlSchwan (changes requested):
Tests are still failing and it seems to be directly related to this PR

lib/Folder/FolderManager.php
```diff
@@ -453,10 +453,16 @@ public function getFoldersForGroups(array $groupIds, int $rootStorageId = 0): ar
...
 			)
-			->where($query->expr()->in('a.group_id', $query->createNamedParameter($groupIds, IQueryBuilder::PARAM_STR_ARRAY)));
+			->where($query->expr()->in('a.group_id', $query->createParameter('groupIds')));
 		$this->joinQueryWithFileCache($query, $rootStorageId);
 
-		$result = $query->executeQuery()->fetchAll();
+        // add chunking because Oracle can't deal with more than 1000 values in an expression list for in queries.
+        $result = [];
+        foreach (array_chunk($groupIds, 1000) as $chunk) {
+            $query->setParameter('groupIds', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
```
CarlSchwan, line 462:
```suggestion
            $query->setParameter('groupIds', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
```

Probably should fix the tests

to-kn:
@CarlSchwan thank you for your feed-back. I did not manage to get the Tests running locally so i was not able to check it beforehand...

to-kn:
@CarlSchwan since tests are running now, i sqashed the commits again. Let me know if anything is missing

CarlSchwan (approved):
LGTM

fschrempf (approved):

PR #1825, opened 2021-12-20
[stable23] Unify folder retrival in commands

CarlSchwan:

Using https://github.com/nextcloud/groupfolders/pull/1825/files?diff=unified&w=1 make the diff easier to read

### DIFF
--- a/lib/Command/ACL.php
+++ b/lib/Command/ACL.php
@@ -22,6 +22,7 @@
 namespace OCA\GroupFolders\Command;
 
 use OC\Core\Command\Base;
+use OCA\GroupFolders\Command\FolderCommand;
 use OCA\GroupFolders\ACL\ACLManagerFactory;
 use OCA\GroupFolders\ACL\Rule;
 use OCA\GroupFolders\ACL\RuleManager;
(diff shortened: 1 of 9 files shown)


come-nc (commented):
The psalm baseline changes are weird, there are duplicated lines. Is there anyway to avoid the new MissingDependency error?

CarlSchwan:
> The psalm baseline changes are weird, there are duplicated lines. Is there anyway to avoid the new MissingDependency error?

Yes we should do something like https://github.com/nextcloud/files_antivirus/pull/218/files#diff-45b5cd912af83b8c2428b3dc1bfcd9e32aa77eec0d68817d056ef9e45e5b6507 with stub implementation for the private API usage we do

come-nc (approved):

CarlSchwan:
For 22 and 21, I wonder if we shouldn't merge the simpler patch https://github.com/nextcloud/groupfolders/pull/1820 and https://github.com/nextcloud/groupfolders/pull/1821. The backporting is getting more and more complicated

come-nc:
> For 22 and 21, I wonder if we shouldn't merge the simpler patch #1820 and #1821. The backporting is getting more and more complicated

I would at least add a is_numeric check, or same check you put in this one with (string) to avoid surprises. Otherwise sure if the backport is too complicated merge the simple version without refactor.

PR #1828, opened 2021-12-20
$folder['permissions'] is always null

CarlSchwan:

### DIFF
--- a/lib/Command/ACL.php
+++ b/lib/Command/ACL.php
@@ -153,7 +153,7 @@ class ACL extends FolderCommand {
 				'/dummy/files/' . $folder['mount_point'],
 				Constants::PERMISSION_ALL,
 				$folder['quota'],
-				$folder['rootCacheEntry'],
+				null,
 				null,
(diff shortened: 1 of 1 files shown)


lib/Command/ACL.php
```diff
@@ -150,7 +151,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 			$mount = $this->mountProvider->getMount(
 				$folder['id'],
 				'/dummy/files/' . $folder['mount_point'],
-				$folder['permissions'],
+				Constants::PERMISSION_ALL,
 				$folder['quota'],
 				$folder['rootCacheEntry'],
```
come-nc, line 156: From what I’ve seen rootCacheEntry is also not present in $folder array

come-nc (changes requested):

lib/Command/ACL.php
CarlSchwan: Good catch, fixed!

come-nc (approved):

PR #1829, opened 2021-12-20
Improve psaml add stub implementation

CarlSchwan:
This allows to now run psalm in many new parts of the codebase (e.g. version backend, trash backend and commands)

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -108,7 +108,13 @@ class TrashBackend implements ITrashBackend {
 		}, $content);
 	}
 
+	/**
+	 * @return void
+	 */
 	public function restoreItem(ITrashItem $item) {
+		if (!($item instanceOf GroupTrashItem)) {
+			throw new \LogicException('Trying to restore normal trash item in group folder trash backend');
+		}
 		$user = $item->getUser();
 		[, $folderId] = explode('/', $item->getTrashPath());
 		$node = $this->getNodeForTrashItem($user, $item);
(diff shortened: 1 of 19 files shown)


lib/Trash/TrashBackend.php
```diff
@@ -224,26 +237,24 @@ private function unwrapJails(IStorage $storage, string $internalPath): array {
...
 		$folders = $this->folderManager->getFoldersForUser($user);
 		$folderIds = array_map(function (array $folder) {
 			return $folder['folder_id'];
 		}, $folders);
 		return in_array($folderId, $folderIds);
 	}
 
-	private function userHasAccessToPath(IUser $user, string $path, $permission = Constants::PERMISSION_READ) {
+	/**
+	 * @psalm-param 1|2|8 $permission
```
come-nc, line 249: Isn’t permission a mask that may be an aggregate of permissions?
Meaning it may be 3 for read&write for instance.

lib/Trash/TrashBackend.php
```diff
@@ -290,6 +304,7 @@ private function getTrashForFolders(IUser $user, array $folders) {
 			$trashFolder = $this->getTrashFolder($folderId);
 			$content = $trashFolder->getDirectoryListing();
 			foreach ($content as $item) {
+                /** @var \OC\Files\Node\Node $item */
```
come-nc, line 307: Indentation seems wrong, unless this is a github display bug.

come-nc (approved):

lib/Trash/TrashBackend.php
CarlSchwan: Psaml auto-fix tried to be clever :) Fixed now

CarlSchwan:
I'll prefer if this get merged after the new year since I will be on vacation starting tomorrow and this needs a bit more testing to make sure that the strong typing doesn't break anything :)

CarlSchwan:
I tested everything again locally and it seems to work

PR #1857, opened 2022-01-04
[stable22] Fix double use of the same variable

CarlSchwan:
Caused by a bad backport in https://github.com/nextcloud/groupfolders/pull/1836


fschrempf (approved):
Nitpick: You've removed the first occurrence here, while for master and NC23, you have removed the second one. Not that it matters, but it might cause conflicts for backports in the future.

acsfer (approved):
Tested 🆗

PR #1874, opened 2022-01-12
Cleanup ExpireGroupVersions and remove not needed arguments

CarlSchwan:
This also fix the following error:

'lib/AppInfo/Application.php on line 151 and exactly 4 expected'

### DIFF
--- a/lib/BackgroundJob/ExpireGroupVersions.php
+++ b/lib/BackgroundJob/ExpireGroupVersions.php
@@ -1,6 +1,9 @@
-<?php declare(strict_types=1);
+<?php
+
+declare(strict_types=1);
 /**
  * @copyright Copyright (c) 2018 Robin Appelman <robin@icewind.nl>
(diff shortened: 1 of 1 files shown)


lib/BackgroundJob/ExpireGroupVersions.php
```diff
@@ -21,47 +24,20 @@
...
-		$this->config = $config;
 	}
 
 	protected function run($argument) {
 		$this->expireManager->expireAll();
-		$backgroundJob = $this->config->getAppValue('files_trashbin', 'background_job_expire_trash', 'yes');
-		if ($backgroundJob === 'no') {
-			return;
-		}
-		$this->trashBackend->expire($this->expiration);
```
artonge, line 65: Why is this not needed anymore? Does `expireAll` run `expire` on the trashbin folder?

CarlSchwan: Because this code is in `lib/BackgroundJob/ExpireGroupTrash.php` now

artonge (approved):

PR #1875, opened 2022-01-12
Update psalm and cs-fixer

CarlSchwan:
<s>The error reported is fixed in https://github.com/nextcloud/groupfolders/pull/1874</s> Rebased


PVince81 (approved):
:+1: nice one!

PR #1887, opened 2022-01-31
Complete getFolder() by adding the manage folder-mappings.

rotdrop:
Signed-off-by: Claus-Justus Heine <himself@claus-justus-heine.de>

### DIFF
--- a/lib/DAV/ACLPlugin.php
+++ b/lib/DAV/ACLPlugin.php
@@ -171,8 +171,7 @@ class ACLPlugin extends ServerPlugin {
 
 		$propFind->handle(self::ACL_ENABLED, function () use ($fileInfo) {
 			$folderId = $this->folderManager->getFolderByPath($fileInfo->getPath());
-			$folder = $this->folderManager->getFolder($folderId, -1);
-			return $folder['acl'];
+			return $this->folderManager->getFolderAclEnabled($folderId);
 		});
 
 		$propFind->handle(self::ACL_CAN_MANAGE, function () use ($fileInfo) {
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -176,6 +176,21 @@ class FolderManager {
 		return $folderMap;
 	}
 
+	/**
+	 * @return array[]
+	 *
+	 * @psalm-return array<int, list<mixed>>
+	 * @throws Exception
+	 */
+	private function getFolderMappings(int $id): array {
+		$query = $this->connection->getQueryBuilder();
+		$query->select('*')
+			->from('group_folders_manage')
+			->where($query->expr()->eq('folder_id', $query->createNamedParameter($id, IQueryBuilder::PARAM_INT)));
+
+		return $query->executeQuery()->fetchAll();
+	}
+
 	private function getManageAcl(array $mappings): array {
 		return array_filter(array_map(function ($entry) {
 			if ($entry['mapping_type'] === 'user') {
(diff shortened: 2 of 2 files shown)


rotdrop:
Hi, I just recently mis-used the routes of the groupfolders app to automatically create a couple of group-shared folders. There I discovered that the getFolder route does not fill the "manage" component as the `FolderManager::getFolder() method does not fill this component. It seem, however, that the get-group-folder-by-id is no where used in the app, so this is not really a bug-fix as this is not really a bug. However, it would be convenient for me if get-folder-by-id would remain and fill the "manage" component.

icewind1991 (approved):
Fine by me

CarlSchwan:
In the current state, this PR will create a new DB request for every dav PROPFIND request but not use the additional info (see lib/DAV/ACLPlugin.php:174). I'm not really a fan of that if this doesn't fix a bug.

Could you make it so that the dav ACLPlugin doesn't call `getFolder` anymore but another method that just returns the ACL or false?

rotdrop:
> In the current state, this PR will create a new DB request for every dav PROPFIND request but not use the additional info (see lib/DAV/ACLPlugin.php:174). I'm not really a fan of that if this doesn't fix a bug.

Oops. I wasn't aware of this. Not good.

> Could you make it so that the dav ACLPlugin doesn't call `getFolder` anymore but another method that just returns the ACL or false?

I'll have a look. Should not be too difficult. Thanks for pointing out this problem!

rotdrop:
> In the current state, this PR will create a new DB request for every dav PROPFIND request but not use the additional info (see lib/DAV/ACLPlugin.php:174). I'm not really a fan of that if this doesn't fix a bug.
> 
> Could you make it so that the dav ACLPlugin doesn't call `getFolder` anymore but another method that just returns the ACL or false?

Ok, I have now a version which does this. Some comments:

- ATM it is "over engineered". I have added two possibility: `getFolderProperty($id, PROPERTY, $rootStorageId = -1)` which also can retrieve the ACL (== ACL is enabled), but also every other single property.
- another minimal `getFolderAclEnabled($id)` version which just returns the "ACL"-flag. This probably should be the one to use
- also in its current state the DAV-plugin's call to getFolder()  (meaning: without this PR) is slightly sub-optimal in that it calls `getFolder($id, rootStorageId: -1)` which results in an unnecessary join with the file-cache tables. So it would probably be beneficial to use something like the `getFolderAclEnabled()` even if this PR is not accepted.

rotdrop:
Just a comment: I seem to have write access and could merge the PR myself, at least the Github UI suggests this. This is probably caused by the fact that was added to the Nextcloud Github-project in the course of some patches to the ldap-code. Maybe my permissions should be revisited.

CarlSchwan:
I would vote for the not "over engineered" solution ;) Let's keep the code base simple and not need to over-optimize the occ commands and admin setting page. The important part is the PROPFIND where we get one join less :)

CarlSchwan:
Regarding the permissions and the fact that you can theoretically merge, this is because you got one approval from someone with write access to the repo. This is expected and this is why the release process moved to the nextcloud-release organization

rotdrop:
> I would vote for the not "over engineered" solution ;) Let's keep the code base simple and not need to over-optimize the occ commands and admin setting page. The important part is the PROPFIND where we get one join less :)

Just squashed and force-pushed the new single commit.

CarlSchwan (approved):
Nice!

PR #1889, opened 2022-02-03
set mount provider in mountpoint

icewind1991:
For https://github.com/nextcloud/server/pull/30978


CarlSchwan:
Should this get merged now?

PR #1891, opened 2022-02-05
Fix more than 1000 files inside a group folder

CarlSchwan:


PVince81 (approved):
:+1: looks great

PR #1915, opened 2022-03-04
Mark background jobs as time sensitive

CarlSchwan:

### DIFF
--- a/lib/BackgroundJob/ExpireGroupTrash.php
+++ b/lib/BackgroundJob/ExpireGroupTrash.php
@@ -50,6 +50,7 @@ class ExpireGroupTrash extends \OC\BackgroundJob\TimedJob {
 		$this->trashBackend = $trashBackend;
 		$this->expiration = $expiration;
 		$this->config = $config;
+		$this->setTimeSensitivity(self::TIME_INSENSITIVE);
 	}
 
 	protected function run($argument) {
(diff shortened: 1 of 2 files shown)


lib/BackgroundJob/ExpireGroupTrash.php
```diff
@@ -50,6 +50,7 @@ public function __construct(
 		$this->trashBackend = $trashBackend;
 		$this->expiration = $expiration;
 		$this->config = $config;
+		$this->setTimeSensitivity(self::TIME_SENSITIVE);
```
PVince81, line 53: this is the default

if you think this should only run at night in the maintenance window, you need to set it to "TIME_INSENSITIVE"

CarlSchwan: 🙃

lib/BackgroundJob/ExpireGroupVersions.php
```diff
@@ -33,6 +33,7 @@ class ExpireGroupVersions extends \OC\BackgroundJob\TimedJob {
 	public function __construct(GroupVersionsExpireManager $expireManager) {
 		// Run once per hour
 		$this->setInterval(60 * 60);
+		$this->setTimeSensitivity(self::TIME_INSENSITIVE);
```
PVince81, line 36: hmm, now I see that we didn't make the regular version expiry time insensitive: https://github.com/nextcloud/server/blob/master/apps/files_versions/lib/BackgroundJob/ExpireVersions.php#L52

and since this was set to run every hour I'm not sure about the impact if we have expiration only running during the night

@nickvergessen what were your thoughts when leaving the version expiration unchanged in core ?

nickvergessen: Well if the job should run once an hour it is not suitable for this. if it runs only in maintenance time, it could run once a day only as well or is it doing junking? then it could run all the time in maintenance window?

I skipped the things in core when the time was not "once a day" as that seemed to not be what we where looking for.

PVince81: alright, thanks for confirming. let's close this

Issue #1744, opened 2021-11-11
Log ACL permission changes to audit log


PVince81:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
ACL permission changes are not written to the audit log. It seems we're already doing that for regular shares and also change of permissions for shares.

**Describe the solution you'd like**
To be able to get a trail of who changed what permissions, we should log ACL permission changes to the audit log.

PVince81:
As far as I understood how the admin_audit app works, this would require triggering hooks from the groupfolders app and catching them in the admin_audit app to log matching messages.

I'm not sure if there's a way to trigger a generic/custom hook to tell the app to log an audit message.

fschrempf:
@PVince81 Would be nice if you could use the issue template next time (I updated your post accordingly), as we are also asking other reporters to use it. Thanks!

joshtrichards:
Fixed in #1917

PR #1917, opened 2022-03-04
Dispatch audit event when changing ACL rules

CarlSchwan:
Implements #1744

### DIFF
--- a/lib/ACL/RuleManager.php
+++ b/lib/ACL/RuleManager.php
@@ -26,18 +26,20 @@ namespace OCA\GroupFolders\ACL;
 use OCA\GroupFolders\ACL\UserMapping\IUserMapping;
 use OCA\GroupFolders\ACL\UserMapping\IUserMappingManager;
 use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IDBConnection;
 use OCP\IUser;
+use OCP\Log\Audit\CriticalActionPerformedEvent;
 
 class RuleManager {
-	/** @var IDBConnection */
-	private $connection;
-	/** @var IUserMappingManager */
-	private $userMappingManager;
+	private IDBConnection $connection;
+	private IUserMappingManager $userMappingManager;
+	private IEventDispatcher $eventDispatcher;
 
-	public function __construct(IDBConnection $connection, IUserMappingManager $userMappingManager) {
+	public function __construct(IDBConnection $connection, IUserMappingManager $userMappingManager, IEventDispatcher $eventDispatcher) {
 		$this->connection = $connection;
 		$this->userMappingManager = $userMappingManager;
+		$this->eventDispatcher = $eventDispatcher;
 	}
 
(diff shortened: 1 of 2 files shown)


lib/ACL/RuleManager.php
```diff
@@ -293,6 +295,20 @@ public function saveRule(Rule $rule): void {
 				->andWhere($query->expr()->eq('mapping_type', $query->createNamedParameter($rule->getUserMapping()->getType())))
 				->andWhere($query->expr()->eq('mapping_id', $query->createNamedParameter($rule->getUserMapping()->getId())));
 			$query->executeStatement();
+
+			if ($rule->getUserMapping()->getType() === 'user') {
+				$logMessage = 'The ACL rule was updated for file/folder with id: "%s" for the user "%s" with permission "%s" and mask "%s"';
```
PVince81, line 300: Maybe change to this so we group the "to" and "for" together
"The ACL rule was updated to permission "%s" and mask "%s" for file/folder with id %s for user "%s""

lib/ACL/RuleManager.php
```diff
@@ -293,6 +295,20 @@ public function saveRule(Rule $rule): void {
...
+			} else {
+				$logMessage = 'The ACL rule was updated for file/folder with id: "%s" for the group "%s" with permission "%s" and mask "%s"';
+			}
+			$params = [
+				'fileId' => $rule->getFileId(),
```
PVince81, line 305: are we mixing parameter order with named parameters ? with associative arrays I'm not sure if the order is guaranteed
is this how we do it in other places, looks weird

lib/ACL/RuleManager.php
```diff
@@ -293,6 +295,20 @@ public function saveRule(Rule $rule): void {
...
+				'user' => $rule->getUserMapping()->getDisplayName() . '(' . $rule->getUserMapping()->getId() . ')',
```
PVince81, line 306: getUserMapping() also for groups ?

lib/ACL/RuleManager.php
```diff
@@ -293,6 +295,26 @@ public function saveRule(Rule $rule): void {
 				->andWhere($query->expr()->eq('mapping_type', $query->createNamedParameter($rule->getUserMapping()->getType())))
 				->andWhere($query->expr()->eq('mapping_id', $query->createNamedParameter($rule->getUserMapping()->getId())));
 			$query->executeStatement();
+
+			if ($rule->getUserMapping()->getType() === 'user') {
+				$logMessage = 'The ACL rule was updated to permission "%s" ans mask "%s" for file/folder with id: "%s" for user "%s"';
```
PVince81, line 300: s/ans/and/ all messages

I'd also remove the colon to make it flow better

PVince81 (approved):
:+1:

would be good to also add a unit test that catches the event and checks at least the params, I think we did this in other places

lib/ACL/RuleManager.php
CarlSchwan: done

PR #1938, opened 2022-03-28
Block sharing root of groupfolder

CarlSchwan:
Currently doesn't work even if when using the debugger the permissions are correctly set. I wonder if the permissions of the root/mount point are defined somewhere else...


PVince81:
superseded by different approach: https://github.com/nextcloud/groupfolders/pull/2013

PVince81:
superseded by different approach: https://github.com/nextcloud/groupfolders/pull/2013

Issue #1760, opened 2021-10-19
Restored (deleted) file is not synced to Linux desktop client.


FreeJulianAssange:
_Should I have posted this at https://github.com/nextcloud/server/issues??_

Yesterday I deleted a file from my NC sync folder on my Linux PC (synced as normal).
Today I restored that deleted file via the Nextcloud Web Files tab/app, however it did not sync the file back to my Linux computer.
The file showed up in the Web UI and also on the android NC app.

I tried to _force sync_ the Linux client without success (it believed it was synced but doesn't resync the restored file).
I _solved_ it by renaming the file on the NC Web/Files tab which triggered the client to sync the file to Linux.

I'm a Linux beginner/noob but looking into the the nextcloud.log file didn't show any errors or exceptions matching the time of the files restoration, but please do let me know if I can check or provide any further information?

**UPDATE**
So its seems I can replicate this issue.
1. Delete file from linux folder (Thunar file browser).
2. Changes are reflected on server and all devices.
3. Restore file from the NC Web interface.
4. All devices, except the Linux client restore the file.
5. Rename file and it syncs as normal.

I also tried moving the file to a new folder, but the Linux NC client reported an error (I couldn't see the full error msg but it mentioned the file Checksum didn't match the file and something about restore.... but message was cut off, I don't know who to find NC Client app logs?).
However all other devices synced the change/move without error and I could see the file & access it from the android NC app.
It seems only renaming the file solves this.

I also noticed the Nextcloud Web/Files tab reports the folder as 0kb in size, after the file is restored, even though the file is 2mb in size.

I've posted some of the NC client logs below, I don't feel comfortable posting the entire logs so I've tried to find the most relevant parts.

In case it helps, this particular folder is shared with a user group and is NOT an E2E folder (E2E is currently disabled as it's not compatible with NC 22.2.0)

--
Linux Mint 20.2
Kernel: Linux 5.11.0-38-generic

Nextcloud Desktop Client
Version 3.3.5 (Linux)

Nextcloud 22.2.0 (Stable)
Ubuntu Server 20.04LTS (LAMP)
PHP7.4

**NC Client App Logs**
```

[...]

PR #1962, opened 2022-04-21
Update parent folders cache when restoring files from trash

mejo-:
Call `Updater->renameFromStorage()` in `TrashBackend->restoreItem()` to
make sure that the file cache of parent folder gets updated.

This fixes a bug where files that got restored to a groupfolder don't
get synced by clients (due to missing update of the parents etag).

Fixes: #1760

Signed-off-by: Jonas <jonas@freesources.org>


CarlSchwan (approved):
Make sense :+1:

juliushaertl (approved):

PR #1996, opened 2022-05-13
specify return type in Rule::jsonSerialize

icewind1991:
otherwise you get an `Error: Return type of Rule::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize()`
error with php 8.1

Signed-off-by: Robin Appelman <robin@icewind.nl>

### DIFF
--- a/lib/ACL/Rule.php
+++ b/lib/ACL/Rule.php
@@ -98,7 +98,7 @@ class Rule implements XmlSerializable, XmlDeserializable, \JsonSerializable {
 		$writer->write($data);
 	}
 
-	public function jsonSerialize() {
+	public function jsonSerialize(): array {
 		return [
 			'mapping' => [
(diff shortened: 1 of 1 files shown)


lib/ACL/Rule.php
```diff
@@ -98,7 +98,7 @@ public function xmlSerialize(Writer $writer) {
 		$writer->write($data);
 	}
 
-	public function jsonSerialize() {
+	public function jsonSerialize(): array {
```
CarlSchwan, line 101:
```suggestion
	#[\ReturnTypeWillChange]
	public function jsonSerialize() {
```

So that it is still compatible with older version

come-nc (approved):
See @CarlSchwan comment if you need to keep PHP<7.4 compatibility.

PVince81:
@CarlSchwan merge ?

PR #2046, opened 2022-08-03
Port licensing information to reuse format

CarlSchwan:
See discussion in https://github.com/nextcloud/server/issues/29723

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,0 +1,16 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: profiler
+Upstream-Contact: Carl Schwan <carl@carlschwan.eu>
+Source: https://github.com/nextcloud/profiler
+
+Files: .github/* .editorconfig babel.config.js .php_cs-fixer.dist.php package-lock.json package.json composer.json composer.lock .gitignore .jshintrc .l10nignore screenshots/* .babelrc.js .eslintrc.js .gitignore .nextcloudignore .tx/config tests/psalm-baseline.xml krankerl.toml postcss.config.js stylelint.config.js tsconfig.json
+Copyright: none
+License: CC0-1.0
+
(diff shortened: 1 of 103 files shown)


README.md
```diff
@@ -1,3 +1,11 @@
+<!--
+SPDX-FileCopyrightText: 2021 Frieder Schrempf
+SPDX-FileCopyrightText: 2021 Carl Schwan <carl@carlschwan.eu>
+SPDX-FileCopyrightText: 2021 Cyrille Bollu
+SPDX-FileCopyrightText: 2021 Pierre Ozoux <pierre@ozoux.net>
+SPDX-FileCopyrightText: 2021 Julius Härtl <jus@bitgrid.net>
+SPDX-License-Identifier: AGPL-3.0-or-later
+-->
```
fschrempf, line 8: Are these tags auto-generated from the git history?
Why are some e-mail addresses missing?
Do we really need copyright and license information in documentation files such as `README.md`? If yes is AGPL the correct license for documentation files or should they be CC0 or something else?

.reuse/dep5
```diff
@@ -0,0 +1,16 @@
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: profiler
```
juliushaertl, line 2:
```suggestion
Upstream-Name: groupfolders
```

.reuse/dep5
```diff
@@ -0,0 +1,16 @@
...
+Upstream-Contact: Carl Schwan <carl@carlschwan.eu>
+Source: https://github.com/nextcloud/profiler
```
juliushaertl, line 4:
```suggestion
Source: https://github.com/nextcloud/groupfolders
```

README.md
juliushaertl: Would also be curious what the way would be to update this automatically from the git history. :)

skjnldsv:
Addressed in subsidiary PR already :)

PR #2057, opened 2022-08-10
enable encryption with system keys for groupfolders

icewind1991:
- [ ] requires https://github.com/nextcloud/server/pull/33500 for the marker interface

It's probably possible to do some backwards compatibility hack to keep compatibility with versions that don't have the marker interface

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -78,6 +78,7 @@ class Application extends App implements IBootstrap {
 			};
 			$config = $c->get(IConfig::class);
 			$allowRootShare = $config->getAppValue('groupfolders', 'allow_root_share', 'true') === 'true';
+			$enableEncryption = $config->getAppValue('groupfolders', 'enable_encryption', 'false') === 'true';
 
 			return new MountProvider(
 				$c->getServer()->getGroupManager(),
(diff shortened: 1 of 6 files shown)


icewind1991:
Changed it so encryption is disabled by default to prevent unexpected behavior change on upgrade

icewind1991:
"backport" is here: https://github.com/nextcloud/groupfolders/pull/2068

lib/AppInfo/Application.php
```diff
@@ -78,6 +78,7 @@ public function register(IRegistrationContext $context): void {
 			};
 			$config = $c->get(IConfig::class);
 			$allowRootShare = $config->getAppValue('groupfolders', 'allow_root_share', 'true') === 'true';
+			$enableEncryption = $config->getAppValue('groupfolders', 'enable_encryption', 'false') === 'true';
```
PVince81, line 81: I'm wondering if we should also bring this to the web UI as it will likely be a more common option than allow_root_share ?

PVince81:
@CarlSchwan can you review this one with the server one also ?

lib/AppInfo/Application.php
kleener: Yes, I am in favor. I would look for it there.

CarlSchwan (approved):

PR #2064, opened 2022-08-17
New fields to use the Groupfolders by a no-admin and API/REST

zak39:
This feature allows adding  2 fields to use the Groupfolders by a no-admin and API/REST.

image

We have :

- The first field :point_right: It gives API/REST and groupfolders administration interface access to all users who are in the groups ;
- The second field :point_right: It gives API/REST access only to all users who are in the groups. But, these users will be access to groupfolders if they are present in the advanced permissions ;

@CarlSchwan, this PR is very useful for the [Workspace](https://github.com/arawa/workspace) app

This PR is a draft for the moment.


CarlSchwan:
Superceded by #2072

PR #2073, opened 2022-08-31
Fix folder with hashtag

CarlSchwan:
Make sure the url is properly encoded otherwise we can't change the ACL
of a folder


PVince81 (approved):
:+1:

but I think this should happen on a higher level, like on "client"

icewind1991 (approved):

PR #2068, opened 2022-08-25
[24] allow enabling encryption for groupfolders

icewind1991:
requires:

- [ ] https://github.com/nextcloud/server/pull/33700

PR #2103, opened 2022-09-27
[stable23] allow enabling encryption for groupfolders

PVince81:
Manual backport of https://github.com/nextcloud/groupfolders/pull/2068 to stable23

Requires https://github.com/nextcloud/server/pull/34102

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -68,7 +68,9 @@ class MountProvider implements IMountProvider {
 	private $mountProviderCollection;
 	private $connection;
 	private $allowRootShare;
-	private bool $enableEncryption;
+
+	/** @var bool */
+	private $enableEncryption;
 
 	public function __construct(
 		IGroupManager $groupProvider,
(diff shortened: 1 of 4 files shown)


PVince81:
I've tested this locally, without encryption and with encryption enabled (need to set the flag with occ config:app:set), and I could see that both encrypted and non-encrypted files could be downloaded.

Newly uploaded files get encrypted.

And the keys are correctly in the system-wide storage.

skjnldsv (approved):
:elephant:

lib/Mount/MountProvider.php
```diff
@@ -68,6 +68,7 @@ class MountProvider implements IMountProvider {
 	private $mountProviderCollection;
 	private $connection;
 	private $allowRootShare;
+	private bool $enableEncryption;
```
blizzz, line 71: typed properties are incompatible with PHP 7.3

blizzz (changes requested):

lib/Mount/MountProvider.php
PVince81: fixed

blizzz (approved):

PR #2114, opened 2022-10-03
return created groupfolder id from create command

### LINKED ISSUE #2104: occ groupfolders:create must output folder id
To make automation possible (ex: ansible), the `occ groupfolders:create` must output the folder id in stdout.
Because the folder id is needed for subsequent group folder commands.

icewind1991:
makes it easier to use in scripting

Fixes #2104


PVince81 (approved):
:+1: thanks

PVince81:
some typing issues in CI

PR #2154, opened 2022-10-26
Move check if node exists earlier

CarlSchwan:
Otherwise we capture the not found exception too late


PVince81 (approved):
:+1: makes sense

PVince81:
backport needed?

Issue #2067, opened 2022-08-21
GroupFolders and BackgroundJob: Failed to remove item from trashbin


MaxiReglisse:
I observe many messages in **nextcloud.log** which relate to the **GroupFolders** application during **background jobs**.

In summary:
```Failed to remove item from trashbin```

More in detail:

```
{"reqId":"WppLnqNirtynluJextbN","level":3,"time":"2022-08-21T14:30:33+00:00","remoteAddr":"","user":"--",
"app":"core","method":"","url":"--","message":"Error while running background job 
(class: OCA\\GroupFolders\\BackgroundJob\\ExpireGroupTrash, arguments: )","userAgent":"--","version":"24.0.3.2",
"exception":{"Exception":"Exception","Message":"Failed to remove item from trashbin","Code":0,
"Trace":[{"file":"/var/www/owncloud/apps/groupfolders/lib/BackgroundJob/ExpireGroupTrash.php","line":58,
"function":"expire","class":"OCA\\GroupFolders\\Trash\\TrashBackend","type":"->"},
{"file":"/var/www/owncloud/lib/public/BackgroundJob/Job.php","line":79,
"function":"run","class":"OCA\\GroupFolders\\BackgroundJob\\ExpireGroupTrash","type":"->"},
{"file":"/var/www/owncloud/lib/public/BackgroundJob/TimedJob.php","line":95,
"function":"execute","class":"OCP\\BackgroundJob\\Job","type":"->"},{"file":"/var/www/owncloud/cron.php",
"line":151,"function":"execute","class":"OCP\\BackgroundJob\\TimedJob","type":"->"}],
"File":"/var/www/owncloud/apps/groupfolders/lib/Trash/TrashBackend.php","Line":394,"message":
"Error while running background job (class: OCA\\GroupFolders\\BackgroundJob\\ExpireGroupTrash, arguments: )",
"CustomMessage":
"Error while running background job (class: OCA\\GroupFolders\\BackgroundJob\\ExpireGroupTrash, arguments: )"}}
```

My config:

Nextcloud version: `24.0.3`
Group Folders: `12.0.1`
Operating system and version: `Debian 11`
Apache:  `2.4.54-1~deb11u1`
PHP version: `7.4.30`
local storage
no encryption

### Steps to reproduce
The problem is traced in the logs every hour during background jobs

no error in /var/log/apache2/error.log

I asked the same question on the community forum, but without success...
[...]

PR #2219, opened 2022-12-29
Log instead of throw for trash item not removable

miaulalala:
Should help with https://github.com/nextcloud/groupfolders/issues/2067

The exception stops execution of the job even though other items might still be in the queue and might not cause issues

Log an error instead of throwing so admins can hande erroring files manually

Signed-off-by: Anna Larch <anna@nextcloud.com>


skjnldsv (approved):

juliushaertl (approved):

juliushaertl:
@miaulalala Please squash the fixup :)

PR #2268, opened 2023-02-17
fix(sidebar): Use icon slot for read-only ACL states

ChristophWurst:
1) Create a user
2) Create a group
3) Create a groupfolder
4) Assign the group
5) Log in as the user
6) Open the sharing sidebar tab of the groupfolder

| Before | After
|---|---|
|  Bildschirmfoto vom 2023-02-17 17-32-29 | Bildschirmfoto vom 2023-02-17 17-28-47 |
|---|---|

### DIFF
--- a/src/components/AclStateButton.vue
+++ b/src/components/AclStateButton.vue
@@ -22,10 +22,14 @@
 <template>
 	<div v-if="readOnly">
 		<NcButton v-if="!isAllowed" v-tooltip="t('groupfolders', 'Denied')">
-			<Cancel :size="16" />
+			<template #icon>
+				<Cancel :size="16" />
+			</template>
 		</NcButton>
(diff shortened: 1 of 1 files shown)


nickvergessen:
If a NcButton does mot have a Text, it needs a title + aria-label 🥲
Browser console should tell that as a warning (I'd really like to see this moved to compile time if possible)

ChristophWurst:
Fair enough. But that is irrelevant to the broken icons, isn't it?

ChristophWurst:
> If a NcButton does mot have a Text, it needs a title + aria-label smiling_face_with_tear

https://github.com/nextcloud/groupfolders/pull/2271 :)

CarlSchwan (approved):

PR #2289, opened 2023-03-06
extend conflict resolution to function with nested groupfolders

icewind1991:
To test

- starting with current master
- create 2 groupfolders `a` and `a/b`, assigning `a` to user `x` and `y` and assigning `a/b` to user `x`
- as user `y` create a folder `b` inside the `a` groupfolder
- as both `x` and `y` upload different files into `a/b` and note that both users see different contents
- apply this patch
- as `x`, open `a` and see that it now contains both `b` and `b (1)`
- as `y`, open `a` and see that `b` has been renamed to `b (1)`

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -125,32 +125,31 @@ class MountProvider implements IMountProvider {
 	public function getMountsForUser(IUser $user, IStorageFactory $loader) {
 		$folders = $this->getFoldersForUser($user);
 
+		// having the folders sorted by mountpoint is required for nested conflict resolution
+		usort($folders, function ($a, $b) {
+			return $a['mount_point'] <=> $b['mount_point'];
+		});
+
 		$mountPoints = array_map(function (array $folder) {
 			return 'files/' . $folder['mount_point'];
 		}, $folders);
-		$conflicts = $this->findConflictsForUser($user, $mountPoints);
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Tested & works

Can't say much about the code changes

lib/Mount/MountProvider.php
```diff
@@ -300,4 +339,57 @@ private function findConflictsForUser(IUser $user, array $mountPoints): array {
...
+			->andWhere($query->expr()->in('path_hash', $query->createParameter('chunk')));
+
+		$paths = [];
+		foreach (array_chunk($pathHashes, 1000) as $chunk) {
+			$query->setParameter('chunk', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
+			foreach ($query->executeQuery()->fetchAll(\PDO::FETCH_COLUMN) as $conflictPath) {
+				[, $folderId, $relativePath] = explode('/', $conflictPath, 3);
+				$conflictsByFolder[(int)$folderId][] = $relativePath;
+			}
+			$paths = array_merge($paths, );
```
ChristophWurst, line 390: this doesn't seem right. otoh $paths isn't read anyway so it doesn't do any harm. but the variable can probably go.

skjnldsv: cc @icewind1991

PR #2327, opened 2023-04-12
allow non-admins to list their own groupfolders

icewind1991:
Allow regular users access to

- `GET apps/groupfolders/folders`
- `GET apps/groupfolders/folders/$folderId`

They only see groupfolders they have access to and only the permissions for groups they belong to

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,3 +119,6 @@ The following REST API's are supported:
     - `mountpoint`: The new name for the folder
 
 For all `POST` calls the required parameters are listed.
+
+Non admins can access the `GET` requests to retrieve info about group folders they have access to.
+Admins can add `applicable=1` as a parameter to the group folder list request to get the same filtered results of only folders they have access to.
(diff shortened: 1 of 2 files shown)


juliushaertl:
Should we somehow also limit the listing for the admin for the cases where this is used in the clients or files app? Otherwise for admins and delegated admins it would list all folders,  also those that are not accessible through the filesystem just for group folder management.

AndyScherzinger:
Would we want/need brute force / throttling on them ? cc @nickvergessen since we lately discussed this and touch endpoints here.

nickvergessen:
I don't have troubles with listing them. Listing does not DDoS/load other users, so that should be fine.

come-nc (approved):
Looks good codewise, not sure about Julius remark.

icewind1991:
Added an option to filter the listing response for admins by adding `applicable=1` to the request params.

PR #2335, opened 2023-04-18
fix text for assigned acl managers not being readable in light mode

icewind1991:
currently it's grey text on grey background in light mode.


provokateurin:
Could we get new releases with this fix?

icewind1991:
fix has been released

PR #2347, opened 2023-04-24
update folder used space after expiring items

icewind1991:
else the "quota exceeded" check will never change.

To test

- create a groupfolder
- upload a >100MB file and a few smaller files
- delete the uploaded files, making sure to delete the large file first
- change the quota of the groupfolder to 100MB
- run `occ groupfolders:expire`
- check which of the trash items were expired.

Currently, because the groupfolder starts with exceeded quota, all items are expired.
With this change, it only expires until the used space drops under the quota.

Also adds some debug logging for the expiry.


juliushaertl (approved):
Code makes senes but haven't tested

susnux (approved):
Tested and works fine :+1:

Issue #2203, opened 2022-12-09
Can not assign `admin` to groupfolder anymore


susnux:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Update to NC 25
2. Add group folder
3. try to assign folder to `admin` group

### Expected behaviour
Can assign `admin` to list of folder owners, like it was possible with NC24 (btw `admin` is member of group folders created before updating to NC25).

### Actual behaviour
Group `admin` is not shown as a option

### Server configuration

**Operating system:** Docker image

**Web server:** Docker image (Apache)

**Database:** MariaDB from Docker Image

**Nextcloud version:** 25.0.2

**Group folders version:** 13.1.0

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated from 24.0.7

**Where did you install Nextcloud from:** Docker

**Are you using external storage, if yes which one:** -

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** LDAP

#### Web server error log
No interesting entries

Sick-Andrew:
We have the same issue, just on a slightly different server.
NC 25, GF 13.1.0

miaulalala:
Hi, you can still assign the groups via occ command:

`occ groupfolders:group YourFolderId admin`

 I'm not sure yet why the dropdown isn't showing the amin group any longer.

christianlupus:
This looks like a regression. There is a [filter function installed](https://github.com/nextcloud/groupfolders/blob/9a3256fa5fb38e219cb31a26c97626d966d1bbd8/src/settings/Api.ts#L53) that filters out the `admin` group. According to the comments that group should have been added at another location in code. Probably that was changed and the adding of the group was overseen.

The quick fix would be to drop the linked line. In the answer from the server, the admin group is present. So, it should work in theory. 

If desired, I can open a short PR with this change.

stepka0815:
I have the same issue, just on a slightly different server with both
NC 25.0.2 & GF 13.1.0 and 
NC 25.0.3 & GF 13.1.1

The problem did not exist with NC24.

Unfortunately, I can only administer NC via browser.

susnux:
[...]

PR #2348, opened 2023-04-25
fix: Include `admin` group in list of all groups

susnux:
* Resolves: #2203 

Currently the `admin` group is **not** automatically added as a member group to a new folder, so users of the `admin` group can not access group folders.

Moreover the `admin` group can currently not be added as a member because it is filtered from the list of available groups.

This PR simply removes that filter, as I think adding `admin` group as a member should be done explicitly, also the other way round we would need to add it to the member groups by default **and** add it to all existing group folders.

### DIFF
--- a/src/settings/Api.ts
+++ b/src/settings/Api.ts
@@ -37,6 +37,7 @@ export interface Folder {
 }
 
 export class Api {
+
 	getUrl(endpoint: string): string {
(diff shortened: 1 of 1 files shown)


miaulalala (approved):
LGTM didn't test

icewind1991 (approved):

Issue #2351, opened 2023-04-26
Extend groupfolders admin UI to be linked to groups and circles


AndyScherzinger:
needs 
- [ ] #2349
- [ ] #2350

Add ability to the admin UI to search and assign circles to groupfolders as well as groups. Circles and its members won't show up for advanced permissions.

Issue #2350, opened 2023-04-26
Extend groupfolders backend to be linked to groups and circles


AndyScherzinger:
Might depend on:
* [ ] #2349

Extend the groupfolders backend to also support circles besides groups being assigned to a groupfolder.

PR #2354, opened 2023-04-28
feat: circles support to groupfolders

ArtificialOwl:
During the configuration process, the front-end can send a circleId instead of a groupId, using the exact same request on the back-end.
The back-end will decide if the id is about a Circle or a Nextcloud Group, by checking that a circle with the id exists or not. If circle does not exists, then the id is the id of a group.

~~The only changes is related to the application returns when listing the groupfolders.
Before this patch, the entry `groups` entry was an array with key=groupId, value=permissions.
Now the value is an array that contains displayName and permissions:~~

Edit: moved new output to a new key to maintain compatibility
```
{
    "id": 1,
    "mount_point": "test_gf001",
    "group": { "admin": 1, "u3HLVORxw8E5IuCZ2UYudIvwDjGIpDa": 7},
    "group_details": {
        "admin": {
            "displayName": "admin",
            "permissions": 1,
            "type": "group"
        },
        "u3HLVORxw8E5IuCZ2UYudIvwDjGIpDa": {
            "displayName": "sub circle",
            "permissions": 7,
            "type": "circle"
        }
    },
    "quota": -3,
    "size": 0,
    "acl": false,
    "manage": []
}
```

When stored in the database, `circle_id` is used for circle, `group_id` for group.

---

- few php8 constructors
- add an endpoint to get available circles to the current viewer. [1]
- add an event listener on circle deletion; to remove any application to all groupfolders based on the singleId of the removed circle.
- change on the applicableMap
- improve the `occ` commands `:group` and `:list` to fit the edit
- update some SQL request to search folders based on the session users' memberships (to circles)
- migration: add `circle_id` to `group_folders_group` and update the unique index

[1] Available Circles:
  - Circles marked as publicly visible.
  - Circles the current viewer is a direct or inherited member of.

Fix https://github.com/nextcloud/groupfolders/issues/2351
Fix https://github.com/nextcloud/groupfolders/issues/2350
Fix https://github.com/nextcloud/groupfolders/issues/2349


icewind1991:
changed the api controller to maintain backwards compatibility by moving the new group info to a new field and putting the old field in the old format

icewind1991:
code looks ok, basic functionality works (you have to add circles from the cli atm, but you can manager permissions from the webui)

ArtificialOwl:
> changed the api controller to maintain backwards compatibility by moving the new group info to a new field and putting the old field in the old format

This will only affect the controller API, the result from occ command (i.e `groupfolders:list --output=json`) wont keep backwards compatibility. 
Are we okay with that ?

tests/psalm-baseline.xml
```diff
@@ -17,14 +17,42 @@
...
+      <code>$queryHelper = $circlesManager->getQueryHelper();</code>
+      <code>$queryHelper = $circlesManager->getQueryHelper();</code>
+      <code>$queryHelper = $circlesManager->getQueryHelper();</code>
+      <code>$circlesManager->startSuperSession();</code>
+      <code>@param CirclesQueryHelper|null $queryHelper</code>
+      <code>$circlesManager->startSuperSession();</code>
+      <code>$circlesManager->getCircle($groupId, $probe);</code>
+      <code>$circlesManager->stopSession();</code>
+      <code>@param CirclesManager|null $circlesManager</code>
+    </UndefinedClass>
```
skjnldsv, line 45: We should fix this instead of committing the new issues :)

skjnldsv (changes requested):

icewind1991:
> > changed the api controller to maintain backwards compatibility by moving the new group info to a new field and putting the old field in the old format
> 
> This will only affect the controller API, the result from occ command (i.e `groupfolders:list --output=json`) wont keep backwards compatibility. Are we okay with that ?

also changed that, good catch

tests/psalm-baseline.xml
icewind1991: This is (mostly?) undefined bits from the circles app, so we would need to add stubs for it

icewind1991 (approved):

skjnldsv (approved):

PR #2342, opened 2023-04-19
Fix GF's ACL's field appearance

Jerome-Herbinet:
Signed-off-by: Jérôme Herbinet <33763786+Jerome-Herbinet@users.noreply.github.com>

fixes https://github.com/nextcloud/groupfolders/issues/2331

Before : 

After :

PR #2341, opened 2023-04-19
Fix quota bar appearance

Jerome-Herbinet:
Signed-off-by: Jérôme Herbinet <33763786+Jerome-Herbinet@users.noreply.github.com>

fixes https://github.com/nextcloud/groupfolders/issues/2332

PR #2355, opened 2023-04-28
UI improvements of the Group Folders' list

Jerome-Herbinet:
Signed-off-by: Jérôme Herbinet <33763786+Jerome-Herbinet@users.noreply.github.com>

For these changes, I was inspired by the appearance of the user list.

## After

Consider this along with the other two (mentioned in the "After" screenshot) : #2341 #2342

### DIFF
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -143,17 +143,25 @@
 					border-bottom: var(--color-border) 1px solid;
 				}
 			}
-			&:hover {
-				background-color: transparent;
-			}
 			&:not(:last-child):hover {
-				box-shadow: 5px 0 0 var(--color-primary-element) inset;
+				td {
+					background-color: var(--color-background-dark);
+					&:first-child {
+						border-radius: 6px 0 0 6px;
+					}
+					&:last-child {
+						border-radius: 0 6px 6px 0;
+					}
+				}
 			}
 			&:last-child {
 				td {
 					padding: 35px 0;
 				}
 			}
+			&:last-child:hover {
+				background-color: transparent;
+			}			
 			td {
 				a {
 					color: var(--color-text-lighter);
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -22,7 +22,7 @@
 		cursor: pointer;
 
 		.sort_arrow {
-			float: right;
+			padding-left: 10px;
 			color: #888;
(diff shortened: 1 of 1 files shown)


jancborchardt (commented):
Very cool! @Jerome-Herbinet I think all the changes are great, would only say we need to adjust the hover effect.

The blue bar on the left doesn't make it very visible, instead we should have the slight grey background, but with rounded corners so it's just like in the action menus. What do you think?

Jerome-Herbinet:
> Very cool! @Jerome-Herbinet I think all the changes are great, would only say we need to adjust the hover effect.
> 
> The blue bar on the left doesn't make it very visible, instead we should have the slight grey background, but with rounded corners so it's just like in the action menus. What do you think?

Done @jancborchardt :-)

Jerome-Herbinet:
> Very cool! Nice work :) Here is the feedback from my side :)
> 
> > Bold like in users list
> > 1px line like in users list
> > Padding top to better separate the GF creation field and list
> > Text color change via #2342
> 
> All of these looks great! One addition to your change: for the group folder name item in the header, change the color to `--color-text-maxcontrast` so that all header items are the same color.
> 
> > Color like in users list
> 
> Here I think we can stick to `--color-main-text` for the items in the list. I would say this information is important enough to be displayed the regular text color. The header text can be `--color-text-maxcontrast` like you changed it :) what do you think @jancborchardt?
> 
> One bigger change over here could be showing this using a similar UI as tags, that could increase readbility, but of course this is a follow up issue. image
> 
> > New hover like in the users list
> 
> Agreed with @jancborchardt about this :)
> 
> > Quota bar display fix
> 
> Commented on the PR #2341 :)

Waiting for @jancborchardt opinion as mentioned about text color.

Jerome-Herbinet:
I forgot to indicate in the "After" screenshot that the sort arrow has been moved, if you have a look at it. 
Initially, in the "Before" screenshot, it's on the right, near the next column, and it's confusing. 
Many times I wondered if the default arrow targets the first column (GF name) or the second one (Groups).

nimishavijay:
> I forgot to indicate in the "After" screenshot that the sort arrow has been moved, if you have a look at it.

I think this is a good improvement :)

juliushaertl (approved):

PR #2375, opened 2023-05-09
Check for folder existence in FolderController

Antreesy:
Based on customer bug report: all folder changes requests return a success, even if the folder doesn't exist.

Added a private method based on already existing code, to return 404 if folder is not exist.
Sorry for my poor PHP-knowledge, i did the best i could :upside_down_face: Please feel free to suggest changes or edit by yourself.

### DIFF
--- a/lib/Controller/FolderController.php
+++ b/lib/Controller/FolderController.php
@@ -145,6 +145,10 @@ class FolderController extends OCSController {
 		if ($storageId === null) {
 			return new DataResponse([], Http::STATUS_NOT_FOUND);
 		}
+		$folder = $this->manager->getFolder($id, $storageId);
+		if ($folder === false) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
 		return null;
 	}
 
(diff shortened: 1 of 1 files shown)


lib/Controller/FolderController.php
```diff
@@ -139,6 +140,14 @@ public function getFolder(int $id): DataResponse {
 		return new DataResponse($this->formatFolder($folder));
 	}
 
+	private function checkFolderExists(int $id): ?DataResponse {
+		$storageId = $this->getRootFolderStorageId();
+		if ($storageId === null) {
+			return new DataResponse([], Http::STATUS_NOT_FOUND);
+		}
+		return null;
```
juliushaertl, line 148: Currently you only get the root folder id which should always be available. We still need to try to fetch the folder with the provided id, the following should work for that:

```suggestion
		$storageId = $this->getRootFolderStorageId();
		if ($storageId === null) {
			return new DataResponse([], Http::STATUS_NOT_FOUND);
		}
		$folder = $this->manager->getFolder($id, $storageId);
		if ($folder === false) {
			return new DataResponse([], Http::STATUS_NOT_FOUND);
		}
		return null;
```

juliushaertl (changes requested):

lib/Controller/FolderController.php
juliushaertl: Other changes seem fine :)

juliushaertl (approved):

PR #2453, opened 2023-07-06
preload the acl rules for the root of the groupfolders

icewind1991:
Since we always need the rules for the groupfolder roots, load the rules for all folders in one go instead of having to do a request for each


juliushaertl (approved):

juliushaertl:
Backports?

PR #2494, opened 2023-07-29
Optimize lib/Mount to PHP8

solracsf:
PHP8 optimizations.

### DIFF
--- a/lib/Mount/CacheRootPermissionsMask.php
+++ b/lib/Mount/CacheRootPermissionsMask.php
@@ -36,11 +36,15 @@ class CacheRootPermissionsMask extends CacheWrapper {
 
 	protected function formatCacheEntry($entry) {
 		$path = $entry['path'];
-		$isRoot = $path === '' || (strpos($path, '__groupfolders') === 0 && count(explode('/', $path)) === 2);
+		$isRoot = $path === '' || (str_starts_with($path, '__groupfolders/') && count(explode('/', $path)) === 2);
+
+		$entry['scan_permissions'] ??= $entry['permissions'];
+
 		if (isset($entry['permissions']) && $isRoot) {
-			$entry['scan_permissions'] = $entry['permissions'];
 			$entry['permissions'] &= $this->mask;
 		}
+
 		return $entry;
 	}
+
 }
--- a/lib/Mount/GroupFolderStorage.php
+++ b/lib/Mount/GroupFolderStorage.php
@@ -52,34 +52,23 @@ class GroupFolderStorage extends Quota {
 
 	public function getOwner($path) {
 		$user = $this->userSession->getUser();
-		if ($user !== null) {
-			return $user->getUID();
-		}
-		return $this->mountOwner !== null ? $this->mountOwner->getUID() : false;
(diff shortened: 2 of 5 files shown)


lib/Mount/CacheRootPermissionsMask.php
```diff
@@ -36,11 +36,15 @@ public function __construct(ICache $cache, int $mask) {
 
 	protected function formatCacheEntry($entry) {
 		$path = $entry['path'];
-		$isRoot = $path === '' || (strpos($path, '__groupfolders') === 0 && count(explode('/', $path)) === 2);
+		$isRoot = $path === '' || (str_starts_with($path, '__groupfolders/') && count(explode('/', $path)) === 2);
+
+		$entry['scan_permissions'] ??= $entry['permissions'];
```
come-nc, line 41: You will get a warning if `$entry['permissions']` is not defined, and this is changing the logic of the function which was only doing that if isRoot is true before?

lib/Mount/GroupFolderStorage.php
```diff
@@ -52,34 +52,23 @@ public function getFolderId(): int {
...
-		}
-		return $storage->scanner;
+		$storage ??= $this;
+
+		$scanner = match (true) {
+			$storage->instanceOfStorage(ObjectStoreStorage::class) => $storage->scanner ??= new ObjectStoreScanner($storage),
+			default => $storage->scanner ??= new Scanner($storage),
+		};
+
+		return $scanner;
```
come-nc, line 72: This is less readable than before I find.

lib/Mount/MountProvider.php
```diff
@@ -126,29 +126,23 @@ public function getFoldersForUser(IUser $user): array {
...
-		}, $foldersWithAcl);
+		$foldersWithAcl = array_filter($folders, fn(array $folder) => $folder['acl']);
+		$aclRootPaths = array_map(fn(array $folder) => $this->getJailPath($folder['folder_id']), $foldersWithAcl);
+
 		$aclManager = $this->aclManagerFactory->getACLManager($user, $this->getRootStorageId());
 		$aclManager->preloadPaths($aclRootPaths);
 
-		return array_values(array_filter(array_map(function ($folder) use ($user, $loader, $conflicts, $aclManager) {
-			// check for existing files in the user home and rename them if needed
+		array_walk($folders, function (&$folder) use ($user, $loader, $conflicts, $aclManager) {
```
come-nc, line 138: Why replace the array_map with an array_walk?

lib/Mount/MountProvider.php
```diff
@@ -188,7 +184,7 @@ private function getCurrentUID(): ?string {
 		}
 
 		$user = $this->userSession->getUser();
-		return $user ? $user->getUID() : null;
+		return $user?->getUID() ?? null;
```
come-nc, line 187:
```suggestion
		return $user?->getUID();
```

come-nc (commented):
All in all, apart from `fn` usage, I’m not really convinced by the changes, a lot of them are sacrifying readability for compactness, especially all the ternary instead of if/else.

solracsf:
After reading it myself i also believe this is less readable...
Closing.

PR #2543, opened 2023-09-06
Adapt version backend to new features

artonge:
Adapt the version backend to the new files versions feature.

## New
+ Support naming a version
+ Support deleting a version
+ Adds DB table to store version infos
+ Fix version creation that was using the current time instead of the version mtime
+ Adapt `LazyFolder` to new `FileInfo` interface https://github.com/nextcloud/server/pull/38150

## Todo
- [x] Test version import from previous versions
- [x] Check tests

Fix https://github.com/nextcloud/server/issues/39138
Needs https://github.com/nextcloud/server/pull/40296

### DIFF
--- a/lib/Migration/Version16000Date20230821085801.php
+++ b/lib/Migration/Version16000Date20230821085801.php
@@ -1,0 +1,86 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Louis Chmn <louis@chmn.me>
+ *
+ * @author Louis Chmn <louis@chmn.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\GroupFolders\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Auto-generated migration step: Please modify to your needs!
+ */
+class Version16000Date20230821085801 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable("group_folders_versions")) {
+			return null;
+		}
+
+		$table = $schema->createTable("group_folders_versions");
+		$table->addColumn('id', Types::BIGINT, [
+			'autoincrement' => true,
+			'notnull' => true,
+			'length' => 20,
+		]);
+		$table->addColumn('file_id', Types::BIGINT, [
+			'notnull' => true,
+			'length' => 20,
+		]);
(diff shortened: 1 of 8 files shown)


AndyScherzinger (approved):

lib/Migration/Version16000Date20230821085801.php
```diff
@@ -0,0 +1,86 @@
...
+		]);
+		$table->addColumn('timestamp', Types::BIGINT, [
+			'notnull' => true,
+			'length' => 20,
+		]);
+		$table->addColumn('size', Types::BIGINT, [
+			'notnull' => true,
+			'length' => 20,
+		]);
+		$table->addColumn('mimetype', Types::BIGINT, [
```
come-nc, line 72: BIGINT for mimetype? :thinking:

lib/Versions/GroupVersion.php
```diff
@@ -46,12 +41,11 @@ public function __construct(
...
-		File $versionFile,
-		int $folderId
+		string $label,
+		private File $versionFile,
+		private int $folderId,
 	) {
-		parent::__construct($timestamp, $revisionId, $name, $size, $mimetype, $path, $sourceFileInfo, $backend, $user);
-		$this->versionFile = $versionFile;
-		$this->folderId = $folderId;
+		parent::__construct($timestamp, $revisionId, $name, $size, $mimetype, $path, $sourceFileInfo, $backend, $user, $label);
```
come-nc, line 48: Both min and max version will need to be 28 then? Is that already the case in info.xml?

lib/Versions/GroupVersionEntity.php
```diff
@@ -0,0 +1,92 @@
...
+
+	public function setLabel(string $label): void {
+		$metadata = $this->getDecodedMetadata();
+		$metadata['label'] = $label;
+		$this->setDecodedMetadata($metadata);
+		$this->markFieldUpdated('metadata');
+	}
+
+	public function getDecodedMetadata(): array {
+		return json_decode($this->metadata ?? '', true) ?? [];
```
come-nc, line 85: It should throw on error as well, no?

lib/Versions/VersionsBackend.php
```diff
@@ -32,66 +35,118 @@
...
+			// Insert the entry in the DB for the current version.
+			$versionEntity = new GroupVersionEntity();
+			$versionEntity->setFileId($file->getId());
+			$versionEntity->setTimestamp($file->getMTime());
+			$versionEntity->setSize($file->getSize());
+			$versionEntity->setMimetype($this->mimeTypeLoader->getId($file->getMimetype()));
+			$versionEntity->setDecodedMetadata([]);
+			$this->groupVersionsMapper->insert($versionEntity);
+
+			// Insert entries in the DB for existing versions.
```
come-nc, line 91: I’m lost, why is that needed outside of migrations?

lib/Versions/VersionsBackend.php
```diff
@@ -32,66 +35,118 @@
...
+				$versionEntity->getTimestamp(),
+				$file->getName(),
+				$versionEntity->getSize(),
+				$this->mimeTypeLoader->getMimetypeById($versionEntity->getMimetype()),
+				$userFolder->getRelativePath($file->getPath()),
+				$file,
+				$this,
+				$user,
+				$versionEntity->getLabel(),
+				$file->getMtime() === $versionEntity->getTimestamp() ? $file : $versionsFolder->get((string)$versionEntity->getTimestamp()),
```
come-nc, line 143: Is there any check to avoid a mess if two versions are created within 1 second?

lib/Versions/VersionsBackend.php
come-nc: Ok, got it, outside of migration the function called before will find versions in DB and this will not run.

lib/Versions/VersionsBackend.php
```diff
@@ -211,4 +267,67 @@ private function getVersionsFolder(int $folderId): Folder {
...
+		$sourceFile = $version->getSourceFile();
+		$mount = $sourceFile->getMountPoint();
+
+		if (!($mount instanceof GroupMountPoint)) {
+			return;
+		}
+
+		$versionsFolder = $this->getVersionsFolder($mount->getFolderId())->get((string)$sourceFile->getId());
+		/** @var Folder $versionsFolder */
+		$versionsFolder->get((string)$version->getTimestamp())->delete();
```
come-nc, line 293:
```suggestion
		$versionsFolder->get((string)$version->getRevisionId())->delete();
```
?
I know it should match the timestamp but…

artonge:
For the backport
- [ ] Bump app version in `info.xml` to trigger the migration

lib/Migration/Version16000Date20230821085801.php
artonge: Because https://github.com/nextcloud/server/blob/6f80ae274bb7795eab9517a4278d394e8da156fb/core/Migrations/Version13000Date20170718121200.php#L164-L168

PR #2560, opened 2023-09-17
pass around preload rules in acl wrapper instead of relying on the cache

icewind1991:
Since the cache is intentionally limited in size (for memory reasons), it doesn't "work" for large folders or when you have a lot of groupfolders with acl enables.

This adds some more manual passing of preloaded rules around to avoid the need to rely on the generic cache in those cases.

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -59,7 +59,7 @@ class ACLManager {
 	 * @param array $paths
 	 * @return (Rule[])[]
 	 */
-	private function getRules(array $paths): array {
+	private function getRules(array $paths, bool $cache = true): array {
 		// beware: adding new rules to the cache besides the cap
 		// might discard former cached entries, so we can't assume they'll stay
 		// cached, so we read everything out initially to be able to return it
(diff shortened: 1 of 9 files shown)


artonge (approved):

lib/ACL/ACLManager.php
```diff
@@ -101,18 +103,33 @@ private function getParents(string $path): array {
...
 
+		return $this->calculatePermissionsForPath($path, $rules);
+	}
+
+	public function getPermissionsForPathFromRules(string $path, array $rules): int {
+		$path = ltrim($path, '/');
+		$parents = $this->getParents($path);
+		// filter to only the rules we care about
+		$rules = $nonCachedPaths = array_intersect_key($rules, array_flip($parents));
+		return $this->calculatePermissionsForPath($path, $rules);
```
come-nc, line 129: `$nonCachedPaths` seems unused?

come-nc: Also this intersect_key suggests that `$rules` is indexed by path, so has string keys, which is not documented as phpdoc. It would help understand the logic if `$rules` and `$parents` type were better defined.

icewind1991: Renamed "parent" to "relevant paths" and added some comments in various places to hopefully make the whole bit easier to understand.

come-nc (approved):

PR #2630, opened 2023-11-23
feat(deps): Add Nextcloud 29 support on master

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>


come-nc (approved):
This adds the version number to package.json, so @icewind1991 or anyone releasing this will need to be aware.

PR #2638, opened 2023-11-27
Return empty array if no user is found

solracsf:
Try to fix https://github.com/nextcloud/groupfolders/issues/2559
Assuming it's OK to return an empty array in this case.

### DIFF
--- a/lib/Versions/VersionsBackend.php
+++ b/lib/Versions/VersionsBackend.php
@@ -23,6 +23,7 @@ declare(strict_types=1);
 
 namespace OCA\GroupFolders\Versions;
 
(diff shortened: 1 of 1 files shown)


artonge:
I think a better fix would be to drop the need to find a user folder.
I drafted something here: https://github.com/nextcloud/groupfolders/pull/2640
But haven't had time to test it.

solracsf:
Makes perfectly sense 👍 Closing this one.

artonge:
@solracsf do you experience the issue yourself? If so, I would ask you to test my solution once I properly tested it myself :).

Issue #11, opened 2017-05-16
Color incompatibility


emresaracoglu:
It seems inconsistent with style. I think, can be integrated with [theme application](https://github.com/nextcloud/server/tree/master/apps/theming)?

![fireshot screen capture 018 - dosyalar - datarals _ verinin guvenli saklama noktasi - datarals_com_apps_files__dir _ fileid 2](https://cloud.githubusercontent.com/assets/297968/26105921/bc3f2056-3a4c-11e7-8c1c-8246c810f6d9.png)

icewind1991:
Fixed with 642833fe8c88b0dfc325e47f1f4f36bbc54921ca

PR #2640, opened 2023-11-27
[stable28] Drop need of user folder

artonge:
The versions are bound to the groupfolder and not the current user. So we can get the `File` object without the current user folder.
This helps in situations where no user is logged in, like background jobs.

Psalm errors already exist in stable28

### DIFF
--- a/lib/Versions/VersionsBackend.php
+++ b/lib/Versions/VersionsBackend.php
@@ -70,10 +70,6 @@ class VersionsBackend implements IVersionBackend, INameableVersionBackend, IDele
 			/** @var Folder $versionsFolder */
 			$versionsFolder = $this->getVersionsFolder($mount->getFolderId())->get((string)$fileInfo->getId());
 
-			$userFolder = $this->rootFolder->getUserFolder($user->getUID());
-			$nodes = $userFolder->getById($fileInfo->getId());
-			$file = array_pop($nodes);
-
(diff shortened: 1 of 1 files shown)


luka-nextcloud:
@artonge  It works well as I checked. Could you please fix the merge conflict?

Chartman123:
@artonge I tried the changes on my NC27 instance. The original error disappeared, however, now I get another one:

`TypeError: OCA\GroupFolders\Versions\GroupVersion::__construct(): Argument #11 ($versionFile) must be of type OCP\Files\File, OC\Files\FileInfo given, called in /var/www/html/custom_apps/groupfolders/lib/Versions/VersionsBackend.php on line 132`

solracsf:
@Chartman123 please post comments at https://github.com/nextcloud/groupfolders/issues/2559 so everybody cand read it. Thanks !

solracsf:
@artonge what about to previously check if the entity does not already exists?
Something like, L88:

```suggestion
if (!$this->groupVersionsMapper->exists($versionEntity) {
	$this->groupVersionsMapper->insert($versionEntity);
}
```

Couldn't it help with https://github.com/nextcloud/groupfolders/issues/2559#issuecomment-1855944032 ?

come-nc (approved):

artonge:
> artonge what about to previously check if the entity does not already exists?

It would hide the reason `getVersionsForFileFromDB` fails :thinking:

solracsf:
Can this be backported to 27?

PR #2642, opened 2023-11-27
don't apply acls when scanning

icewind1991:
we want to scan the full storage, even if the current user can't see all the files


come-nc (approved):

solracsf:
This is causing issues with encryption: https://github.com/nextcloud/groupfolders/issues/2793

PR #2631, opened 2023-11-23
Fix ACL rules for trashed subfolders from groupfolders

come-nc:
Cleaner version of https://github.com/nextcloud/groupfolders/pull/2629

PR #2644, opened 2023-11-28
Add a listener to update trashed items when parent is renamed

come-nc:
Make sure that when renaming a folder in a groupfolder items in the trash get their original_location field updated, so that they get restored in the right place later, and ACLs with #2631 are still correctly applied.

Follow-ups:
- Do something sensible when folder is moved out of groupfolders (move content to user trash I suppose?)

### DIFF
--- a/lib/Listeners/NodeRenamedListener.php
+++ b/lib/Listeners/NodeRenamedListener.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Côme Chilliet <come.chilliet@nextcloud.com>
+ *
+ * @author Côme Chilliet <come.chilliet@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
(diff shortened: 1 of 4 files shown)


lib/Listeners/NodeRenamedListener.php
```diff
@@ -0,0 +1,67 @@
...
+
+		if (($target instanceof Folder) &&
+			$sourceStorage->instanceOfStorage(GroupFolderStorage::class) &&
+			$targetStorage->instanceOfStorage(GroupFolderStorage::class)) {
+			// Get internal path on parent to avoid NotFoundException
+			$sourcePath = $source->getParent()->getInternalPath();
+			if ($sourcePath !== '') {
+				$sourcePath .= '/';
+			}
+			$sourcePath .= $source->getName();
```
artonge, line 62: Is `$sourcePath .= $source->getInternalPath();` impossible?

lib/Trash/TrashManager.php
```diff
@@ -90,4 +90,30 @@ public function emptyTrashbin(int $folderId): void {
...
 	}
+
+	public function updateTrashedChildren(int $fromFolderId, int $toFolderId, string $fromLocation, string $toLocation): void {
+		// Update deep children
+		$query = $this->connection->getQueryBuilder();
+		$fun = $query->func();
+		$sourceLength = mb_strlen($fromLocation);
+		$newPathFunction = $fun->concat(
+			$query->createNamedParameter($toLocation),
+			$fun->substring('original_location', $query->createNamedParameter($sourceLength + 1, IQueryBuilder::PARAM_INT))// +1 for the leading slash
```
artonge, line 101: No?

```suggestion
			$fun->substring('original_location', $query->createNamedParameter($sourceLength + 1, IQueryBuilder::PARAM_INT))// +1 for the ending slash
```

artonge (approved):

lib/Listeners/NodeRenamedListener.php
come-nc: Yes, because $source is a NonExistingFile so it triggers a NotFoundException.

Issue #2637, opened 2023-11-27
ACL permissions are not shown in Files sidebar


Antreesy:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

Javascript code fails here with `client === undefined`:
https://github.com/nextcloud/groupfolders/blob/e601d498ad7d497424ae5ea0807db7f95aee72be/src/client.js#L190

### Steps to reproduce
1. Create a groupfolder
2. Enable advanced permissions
3. Open Sharing tab in Files sidebar

### Expected behaviour

Table for setting ACL should be visible to user (screenshot from NC v27, groupfolders 15.3.1, console errors unrelated)
![image](https://github.com/nextcloud/groupfolders/assets/93392545/f0105b73-278f-47ba-8780-b829f4086e7d)

client object is
```
{
    "_root": "/remote.php/dav/files/admin",
    "_host": "stable27.local",
    "_defaultHeaders": {
        "X-Requested-With": "XMLHttpRequest",
        "requesttoken": "token"
    },
    "_baseUrl": "https://stable27.local/remote.php/dav/files/admin",
    "_client": {..},
    "_fileInfoParsers": [..],
    "_propfindProperties": [..]
}
```

### Actual behaviour

Table is missing due to JS failed (screenshot from NC v28, groupfolders 16.0.0)
![image](https://github.com/nextcloud/groupfolders/assets/93392545/37b12c19-3f65-467a-8121-162716ebb3b7)

### Server configuration

**Nextcloud version:** 28+

**Group folders version:** 16.0.0

#### Browser log
<details>
<summary>Browser log</summary>
```
vue.runtime.esm.js:3049 TypeError: Cannot read properties of undefined (reading 'getFileInfo')
    at e.value (client.js:190:17)
    at o.loadAcls (SharingSidebarView.vue:247:1)
    at o.beforeMount (SharingSidebarView.vue:240:1)
    at bn (vue.runtime.esm.js:3017:61)
    at pa (vue.runtime.esm.js:4031:13)
    at e (vue.runtime.esm.js:3854:5)
    at ii.$mount (vue.runtime.esm.js:8772:12)
    at init (vue.runtime.esm.js:4411:19)
[...]

PR #2660, opened 2023-12-08
fix(ACL): don't put inherited ACL permissions in the propPatch request payload

### LINKED ISSUE #2659: Inherited ACL permissions for groups in subfolder converted to local after adding new ACL rules
### How to use GitHub

#### 1. Configure group folder (Root) for groups A, B with default permissions and ACL enabled. Set ACL for group A to 'deny',  to group B 'allow' (via OCC or share sidebar)

`occ groupfolders:list`

| Folder Id | Name | Groups                              | Quota     | Size | Advanced Permissions | Manage advanced permissions |
|-----------|------|-------------------------------------|-----------|------|----------------------|-----------------------------|
| 1         | Root | admin: read, write, share, delete   | Unlimited | 0 B  | Enabled              | admin (group)               |
|           |      | Group A: read, write, share, delete |           |      |                      |                             |
|           |      | Group B: read, write, share, delete |           |      |                      |                             |

`occ groupfolders:permissions 1`

| Path | User/Group     | Permissions                             |
|------|----------------|-----------------------------------------|
| /    | group: Group A | -read, -write, -create, -delete, -share |
|      | group: Group B |                                         |

<details>

image

</details>

#### 2. Create a subfolder, ACL permissions from Root should be inherited (can't edit)

<details>

image

</details>

#### 3. Add a new ACL rule for user/group for subfolder

<details>

User | Group 
-- | --
image | image
image | image

</details>

#### 4. Reload page or re-fetch sidebar data, see that inherited permission were converted to local

User | Group 
-- | --
image | image

#### 5. After removing that user from ACL, group permissions remains local

User | Group 
-- | --
image | image

### Expected behaviour
Tell us what should happen

### Server configuration
**Nextcloud version:** 26.0.8
**Group folders version:** 14.0.5

Antreesy:
* Fix #2659
* Including inherited permissions in `client.propPatch` request makes them local

:warning: As https://github.com/nextcloud/groupfolders/issues/2637 is not solved yet, tested it on stable26 and stable27 only

Before | After
-- | --
create ACL | image
image | image
remove ACL | image
image | image

### DIFF
--- a/src/components/SharingSidebarView.vue
+++ b/src/components/SharingSidebarView.vue
@@ -309,7 +309,7 @@ export default {
 			const rule = new Rule()
 			rule.fromValues(option.type, option.id, option.displayname, 0b00000, 0b11111)
 			this.list.push(rule)
-			client.propPatch(this.model, this.list).then(() => {
+			client.propPatch(this.model, this.list.filter(rule => !rule.inherited)).then(() => {
 				this.showAclCreate = false
 			})
 		},
@@ -319,7 +319,7 @@ export default {
 			if (index > -1) {
 				list.splice(index, 1)
 			}
-			client.propPatch(this.model, list).then(() => {
+			client.propPatch(this.model, list.filter(rule => !rule.inherited)).then(() => {
 				this.list.splice(index, 1)
 				const inheritedAcl = this.inheritedAclsById[rule.getUniqueMappingIdentifier()]
 				if (inheritedAcl != null) {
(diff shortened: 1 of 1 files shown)


Antreesy:
> it will need the UI to work as well I suppose.

Indeed, that would require sidebar view to test it. Now, with cypress we do everything with occ commands, one rule at the time, so impoossible to reproduce this error.

ATM web interface sends all rules together, which are handled by ACLPlugin propPatch method:
https://github.com/nextcloud/groupfolders/blob/0026135c7d58bbabbee7909112faf76b27e54636/lib/DAV/ACLPlugin.php#L175

As seen from there, it doesn't divide coming rules to inherited and local, consider them all as local, so we shouldn't pass inherited unless they haven't been changed.

GretaD (approved):
code looks good but i didnt test it

mejo- (approved):
Please disregard my last comment. For some reason I didn't run the code with your commit. Now that I fixed this, it works as expected :tada:

PR #2669, opened 2023-12-13
Fix NoUserException

solracsf:
Fix https://github.com/nextcloud/server/issues/19647
Fix https://github.com/nextcloud/groupfolders/issues/2559


artonge:
What about https://github.com/nextcloud/groupfolders/pull/2640?

solracsf:
I need coffee ☕ 🙈

PR #2703, opened 2023-12-26
Fix display of files/folders in scan command

SystemKeeper:
Was reported at https://help.nextcloud.com/t/occ-groupfolders-scan-files-and-folders-mixed-up/170749/

The header looks like 
https://github.com/nextcloud/groupfolders/blob/2aff6aaeb4dc615ea2462ede1fe1e4b5aee3f505/lib/Command/Scan.php#L121-L123


come-nc:
This was already fixed in https://github.com/nextcloud/groupfolders/pull/2570 and you are breaking it again, I think?

PR #2720, opened 2024-01-05
fix: duplicate entry for key 'gf_versions_uniq_index'

luka-nextcloud:


luka-nextcloud:
> This does not explain why we do find 0 versions [here](https://github.com/nextcloud/groupfolders/pull/2720/files#diff-d6dddc683b04d4c4f0118e29e9b81d4fb180ae40c4f31beb784c9d3a144712a9R78), but then fail to create a version.
> 
> Have you been able to reproduce the issue?

Here's my case to reproduce this issue:

1. Create a group folder A
2. Create a group folder B
3. Create a file called Test.md in group folder A
4. Edit file Test.md and save
5. Move file Test.md to group folder B
6. Run command occ groupfolders:expire -v

miaulalala:
> This does not explain why we do find 0 versions [here](https://github.com/nextcloud/groupfolders/pull/2720/files#diff-d6dddc683b04d4c4f0118e29e9b81d4fb180ae40c4f31beb784c9d3a144712a9R78), but then fail to create a version.
> 
> Have you been able to reproduce the issue?

Read- write split? If it's a clustered setup there's no guarantee a write will have propagated to all nodes by the time we read, no?

luka-nextcloud:
This case might occur, but only if group version records were not created (or already deleted) somehow. The group version record should be created right after a file updated.

artonge:
I found the issue. The version backend is given a broken fileinfo in the case the Luka described. This leads to an incorrect query which returns no results, and explain that we then fail to insert an entry as it already exists.

Closing in favor of https://github.com/nextcloud/groupfolders/pull/2758

PR #2736, opened 2024-01-09
restore acl ui in sharing tab

icewind1991:
Fixes https://github.com/nextcloud/groupfolders/issues/2637

This can probably use some larger rework to make thing modern™ but for now this seems to fix things

PR #2746, opened 2024-01-11
[28] restore acl ui in sharing tab

icewind1991:
Backport of https://github.com/nextcloud/groupfolders/pull/2736


Antreesy (approved):
Smoke tested, opening and working in general
Maybe it requires some thorough check due to bumping that much libraries in once (checked nc/vue, changes seem ok between 7.12.6 => 8.4.0)

PR #2758, opened 2024-01-16
Handle missing disk versions

artonge:
This PR is a palliative upon groupfolders failing to move versions correctly when a file is moved across storage.

1. Delete versions folder when we detect that the original file in the group folder is not there anymore.(`GroupVersionsExpireManager.php`)
2. Create the version folder before reading it (`VersionsBackend.php`).
3. Delete version entity in the DB if the disk counterpart does not exist (`VersionsBackend.php`).

Fix https://github.com/nextcloud/groupfolders/issues/2559

### DIFF
--- a/lib/Versions/GroupVersionsExpireManager.php
+++ b/lib/Versions/GroupVersionsExpireManager.php
@@ -23,13 +23,14 @@ declare(strict_types=1);
 
 namespace OCA\GroupFolders\Versions;
 
-use OC\Files\FileInfo;
 use OC\Files\View;
 use OC\Hooks\BasicEmitter;
 use OC\User\User;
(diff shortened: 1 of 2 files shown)


skjnldsv (approved):

EDVB-Paulick:
When is this patch available in 28? We really need it.

Cheers

artonge:
> When is this patch available in 28? We really need it.

If this is important to you, please get a subscription at Nextcloud :).

EDVB-Paulick:
> > When is this patch available in 28? We really need it.
> 
> If this is important to you, please get a subscription at Nextcloud :).

You are right, we are working on this issue internally ... ;-)

PR #2797, opened 2024-02-02
Revert: don't apply acls when scanning

icewind1991:
The storage wrapper shenanigans break other wrappers like encryption

reverts: https://github.com/nextcloud/groupfolders/pull/2642

I've also tried to create a test case for the original "fix" but I can't seem to reproduce the issue anymore.

### DIFF
--- a/tests/ACL/ACLScannerTest.php
+++ b/tests/ACL/ACLScannerTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace OCA\groupfolders\tests\ACL;
+
+use OC\Files\Storage\Temporary;
+use OCA\GroupFolders\ACL\ACLManager;
+use OCA\GroupFolders\ACL\ACLStorageWrapper;
+use OCP\Constants;
+use Test\TestCase;
+
+/**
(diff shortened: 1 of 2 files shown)


tests/ACL/ACLScannerTest.php
```diff
@@ -0,0 +1,60 @@
...
+namespace OCA\groupfolders\tests\ACL;
+
+use OC\Files\Storage\Temporary;
+use OCA\GroupFolders\ACL\ACLManager;
+use OCA\GroupFolders\ACL\ACLStorageWrapper;
+use OCP\Constants;
+use Test\TestCase;
+
+/**
+ * @group DB
```
come-nc, line 12: Why group DB?

come-nc (approved):

tests/ACL/ACLScannerTest.php
icewind1991: the test users cache so it needs db access

come-nc:
@icewind1991 This was not backported, it should be right?

PR #2833, opened 2024-02-26
fix: mask out rule permissions before merging permissions

icewind1991:
To test:

- Create group1 and group2
- Create user1 and add it to group1 and group2
- Create a groupfolder with access to group1 and group2 and acl enabled
- On the groupfolder, set all permissions to allow for group1, and all except read to deny for group2
- Create a subfolder `folder`
- On the subfolder, set all to deny for group2.
- `occ groupfolders:permissions 1` should list the following permissions

```
+--------+---------------+-----------------------------------------+
| Path   | User/Group    | Permissions                             |
+--------+---------------+-----------------------------------------+
| /      | group: group2 | -write, -create, -delete, -share        |
|        | group: group1 | +read, +write, +create, +delete, +share |
| folder | group: group2 | -read, -write, -create, -delete, -share |
+--------+---------------+-----------------------------------------+
```

User1 has no access to the subfolder, as it is denied through group2

- On folder, set all share permissions to deny and the rest to inherit
- `occ groupfolders:permissions 1` should list the following permissions

```
+--------+---------------+-----------------------------------------+
| Path   | User/Group    | Permissions                             |
+--------+---------------+-----------------------------------------+
| /      | group: group2 | -write, -create, -delete, -share        |
|        | group: group1 | +read, +write, +create, +delete, +share |
| folder | group: group1 | -share                                  |
|        | group: group2 | -read, -write, -create, -delete, -share |
+--------+---------------+-----------------------------------------+
```

On master: user1 now suddenly has access to the folder, even though there should be no new allow permissions for them.

With this PR his permissions are still denied.

### DIFF
--- a/lib/ACL/Rule.php
+++ b/lib/ACL/Rule.php
@@ -51,7 +51,7 @@ class Rule implements XmlSerializable, XmlDeserializable, \JsonSerializable {
 		$this->userMapping = $userMapping;
 		$this->fileId = $fileId;
 		$this->mask = $mask;
-		$this->permissions = $permissions;
+		$this->permissions = $permissions & $mask;
 	}
 
 	public function getUserMapping(): IUserMapping {
--- a/tests/ACL/ACLCacheWrapperTest.php
+++ b/tests/ACL/ACLCacheWrapperTest.php
@@ -28,9 +28,11 @@ use OCA\GroupFolders\ACL\ACLCacheWrapper;
 use OCA\GroupFolders\ACL\ACLManager;
 use OCP\Constants;
 use OCP\Files\Cache\ICache;
-use OCP\IDBConnection;
 use Test\TestCase;
 
+/**
+ * @group DB
+ */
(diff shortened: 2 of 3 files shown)


lib/ACL/Rule.php
```diff
@@ -145,7 +145,7 @@ public static function mergeRules(array $rules): Rule {
 		}, 0);
 		// or'ing the permissions combines them with allow overwriting deny
 		$permissions = array_reduce($rules, function (int $permissions, Rule $rule) {
-			return $permissions | $rule->getPermissions();
+			return $permissions | ($rule->getPermissions() & $rule->getMask());
```
come-nc, line 148: Reading comments on the constructor, should the mask be applied to permissions in constructor to be safe? It states that `$this->permissions` is supposed to only apply to non-inherited values, so it should only allow it to contain permission for those, no?

come-nc (approved):

PR #2856, opened 2024-03-12
fix: fix fully inherited acl state not being show as inherited

icewind1991:
Fix the ACL state button for inherited rules not being shown as inherited.


come-nc (approved):
:elephant:

Issue #2812, opened 2024-02-09
"Deny" Group permission prevails over "allow" User permission


slaver666:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. User has extended permission to the root folder. Group has read permissions to the root folder. 
2. User creates a folder inside root folder. The child folder inherits root folder permissions.
3. User removes Group read permission to the child folder.  

### Expected behaviour
User must have all permissions to the folder.

### Actual behaviour
User even doesn't see the folder any more.

### Server configuration

**Operating system:**
Ubuntu Linux 22.04.3 LTS

**Web server:**
Nginx 1.25.3

**Database:**
PostgreSQL 15.5

**PHP version:**
8.2

**Nextcloud version:** (see Nextcloud admin page)
28.0.2

**Group folders version:**
16.0.3

**Updated from an older Nextcloud/ownCloud or fresh install:**
updated from 27

**Where did you install Nextcloud from:**
community edition from official site

**Are you using external storage, if yes which one:** local/s3/smb/sftp/...
no

**Are you using encryption:** yes/no
no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/Saml/...
Active Directory

### Client configuration
**Browser:**
Google Chrome, Microsoft Edge

**Operating system:**
Windows 10, Ubuntu Linux 22

### Logs
Can't see any log entries belonged to Group folders app

#### Web server error log
<details>
<summary>Web server error log</summary>

```
No errors in web server log
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
[Uploading nextcloud.log…]()


```
</details>

#### Browser log
<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```
</details>

slaver666:
[nextcloud.log](https://github.com/nextcloud/groupfolders/files/14220892/nextcloud.log)

Jerome-Herbinet:
Hi @slaver666, 
[...]

Issue #598, opened 2019-09-10
Colliding ACL rules when user is in multiple groups


SimJoSt:
We use ACLs for a group folder. This folder is available to multiple groups. Some users are in multiple of those groups.

Some folders have ACL permissions set to prevent one user group to access it, while the other group should still be able to access it, as inherited from the folder above it.  
The users in multiple of the groups now don't have access to this specific folder.

Do this permissions work in an "and" or an "or" way? And can the behaviour be changed?

icewind1991:
If a user has multiple ACL rules matching for a single folder they work in an "or" way

ChrissW-R1:
> If a user has multiple ACL rules matching for a single folder they work in an "or" way (so "allow" overwrites "deny")

That is not right. We have a setting, in which the user is in different groups, with colliding permissions. And the users don't have allowed permissions.


We use this structure:
root
├─┬─ company_a
│   ├─ department_a1
│   └─ department_a2
├─┬─ company_b
│   ├─ department_b1
│   ├─ department_b1
│   └─ department_b3
└─┬─ company_c
     └─ department_c1

Everyone employee should be able to read only everything in his company share. But rights to write should only enabled for the department folder(s), of the assigned department(s) the user is a member of.

![image](https://user-images.githubusercontent.com/1086210/73986161-376dd900-493d-11ea-91b7-a7979b317bf4.png)


Sure it is possible, that we configured something wrong, but we have no idea what. Any tips?

putt1ck:
I've just worked through this for a customer so hopefully this will help:

at groupfolder level grant to the relevant groups the maximum permission the organisation would ever want members of that group to have in the groupfolder (where groupfolder means everything within it, whether subfolders or files, no matter how many levels of subfolders are intended);
with the share tab of the groupfolder (as a user in a group with permission to use advanced permissions) add any additional restrictions you want at that level e.g. in a scenario for company information you probably want only a few individuals to create new subfolders, so you grant them those permissions and then add rules for the groups to restrict to read only;
then on the desired subfolders add rules to grant back permissions to users or groups as needed and so on down the tree.

NB when a user is in more than one group where one has more relaxed permissions then you need a rule for both groups.

fschrempf:
It sounds like this is the problem also described in #1212. A potential fix is in #1654. Please confirm that this would fix the issue. If possible, test/review the mentioned PR.

fschrempf:
[...]

PR #2870, opened 2024-03-18
feat: add option to resolve inherited option per-user instead of per-path

icewind1991:
Instead of always have subfolder rules overwrite parent folder rules, first resolve the inherited rules for each user/group before combining the rules from the different users/groups.

With the default behavior, if any rule applicable to the user denies permission for a subfolder, the permission will be denied, even if the user is member of a group that has "allow" permissions.

With the new behavior, if any group the user is a member for has "allow" permissions, the user has access.

To test:

1. create groups `group1` and `group2`
2. create `user` test in `group1` and `group2`
3. create groupfolder `gf` accessible to `group1` and `group2` an ACLs enabled
4. create folder `gf/a` and create a rule on it for `group1` giving full permissions
5. create folder `gf/a/b` and create a rule on it for `group2`, denying read permissions
6. with default behavior `user` won't have access to `gf/a/b` because of the rule from 5.
7. change to the new behavior with `occ config:app:set groupfolders acl-inherit-per-user --value true`
8. now `user` does have access to `gf/a/b` because `group1` has access, users only in `group2` won't have access

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -35,11 +35,12 @@ class ACLManager {
 	private $rootFolderProvider;
 
 	public function __construct(
-		private RuleManager $ruleManager,
+		private RuleManager  $ruleManager,
 		private TrashManager $trashManager,
-		private IUser $user,
-		callable $rootFolderProvider,
-		private ?int $rootStorageId = null,
+		private IUser        $user,
+		callable             $rootFolderProvider,
+		private ?int         $rootStorageId = null,
+		private bool         $inheritMergePerUser = false,
 	) {
 		$this->ruleCache = new CappedMemoryCache();
 		$this->rootFolderProvider = $rootFolderProvider;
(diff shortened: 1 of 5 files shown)


lib/ACL/ACLManager.php
```diff
@@ -168,12 +169,32 @@ public function getPermissionsForPathFromRules(string $path, array $rules): int
...
+					$key = $mapping->getType() . '/' . $mapping->getId();
+					if (!isset($rulesPerMapping[$key])) {
+						$rulesPerMapping[$key] = Rule::defaultRule();
+					}
+
+					$rulesPerMapping[$key]->applyRule($rule);
+				}
+			}
+
+			$mergedRule = Rule::mergeRules($rulesPerMapping);
```
artonge, line 188: Here we would only have something like:

```php
$rulesPerMapping = [
    'group/gid1' => Rule,
    'group/gid2' => Rule,
    'user/uid' => Rule,
];
```

Where uid is the user id of the current user, and where there can be multiple gid depending on the user's groups, right?

artonge (commented):
Can you confirm that the goal is to have a switch to toggle between:
1. Restrict down the user based on the rules
2. Give as many permissions to the user based on the rules.

lib/ACL/ACLManager.php
icewind1991: yes

artonge (approved):
Approving to get it merged, but I would like to properly understand the logic:

Considering the following hierarchy: `/a/b`

Without `inheritMergePerUser` (old): merge(`b/g1`, `b/g2`)->apply(merge(`a/g1`, `a/g2`))
With `inheritMergePerUser` (new): merge(`b/g1`->apply(`a/g1`), `b/g2`->apply(`a/g2`))?

`applyPermissions`: if the rule does not have a value for a permission, use the provided one
`mergeRules`: do an "or" with both permissions set

| Folder     | Read  | Update | Share | Delete |
| ---------- | :---: | :----: | :---: | :----: |
| a/g1       |   1   |   1    |   1   |   1    |
| a/g2       |   -   |   -    |   -   |   -    |
| b/g1       |   -   |   -    |   -   |   -    |
| b/g2       |   0   |   -    |   -   |   -    |
| b/u1 (old) |   0   |   1    |   1   |   1    |
| b/u1 (new) |   1   |   1    |   1   |   1    |

If that's correct, then perfect :).
Do you think it would make sense to add a comment with the above?

icewind1991:
Yes, I've added a comment, please let me know if that explanation is clear enough

lib/ACL/ACLManager.php
```diff
@@ -168,12 +169,49 @@ public function getPermissionsForPathFromRules(string $path, array $rules): int
...
+		// | a/b: g2     | 0    | -      | -     | -      |
+		// |-------------|------|--------|-------|--------|
+		//
+		// and a user that is a member of g1 and g2
+		//
+		// Without `inheritMergePerUser` the user will not have access to `a/b`
+		// as the merged rules for `a/b` ("-read,-delete") will overwrite the merged for `a` ("+read,+write+share+delete")
+		//
+		// With b`inheritMergePerUser` the user will have access to `a`
+		// a the applied rules for `g1` ("+read,+write+share") merges with the applied rules for `g2` ("-read")
```
come-nc, line 188:
```suggestion
		// With `inheritMergePerUser` the user will have access to `a/b`
		// as the applied rules for `g1` ("+read,+write+share") merges with the applied rules for `g2` ("-read")
```

fschrempf:
It looks like this would solve or at least affect a few open issues (#2812, #1212, #598, etc.). Are there any plans to reference this new feature there to let people know and maybe also have this mentioned in the README?

thorsten-schwartz:
Hi all,
Shouldn't the correct occ be with type=boolean:

occ config:app:set groupfolders acl-inherit-per-user --value="true" --type=boolean

In the documented way value will be a string; that's not nice, is it?

PR #2872, opened 2024-03-19
fix: Set mount point target user as storage owner

juliushaertl:
Requires https://github.com/nextcloud/server/pull/44294

This will avoid that calling `getUserFolder($user)->getByid($id)` will fail on paths that have no share permission without a user session.

Fix https://github.com/ONLYOFFICE/onlyoffice-nextcloud/issues/900


icewind1991 (approved):

juliushaertl:
Failure is unrelated (also happens on other prs)

PR #2916, opened 2024-04-09
fix: fix opening acl create select

icewind1991:
The old way broke when the select was moved to `NcSelect`

### DIFF
--- a/src/components/SharingSidebarView.vue
+++ b/src/components/SharingSidebarView.vue
@@ -305,9 +305,11 @@ export default {
 		},
 		toggleAclCreate() {
 			this.showAclCreate = !this.showAclCreate
-			Vue.nextTick(() => {
-				this.$refs.select.$el.focus()
-			})
+			if (this.showAclCreate) {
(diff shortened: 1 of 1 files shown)


artonge (approved):

src/components/SharingSidebarView.vue
```diff
@@ -305,9 +305,12 @@ export default {
 		},
 		toggleAclCreate() {
 			this.showAclCreate = !this.showAclCreate
-			Vue.nextTick(() => {
-				this.$refs.select.$el.focus()
-			})
+			if (this.showAclCreate) {
+				Vue.nextTick(() => {
+					// hacky way to get the Vue-Select instance from the NcSelect
+					this.$refs.select._vnode.componentInstance.open = true
```
Pytal, line 311:
```suggestion
					this.$refs.select.$el.querySelector('input').focus()
```
This way it will have keyboard focus like before + open the dropdown

Pytal (changes requested):

Pytal (approved):

PR #2951, opened 2024-05-06
fix: Adapt to constructor change of TrashItem in files_trashbin

come-nc:
Ideally we should store the information of who deleted the item and pass it, but for now just pass null to avoid a crash.

This should fix cypress CI.

Related to: https://github.com/nextcloud/server/pull/44643

### DIFF
--- a/lib/Trash/GroupTrashItem.php
+++ b/lib/Trash/GroupTrashItem.php
@@ -39,7 +39,7 @@ class GroupTrashItem extends TrashItem {
 		IUser $user,
 		string $mountPoint
 	) {
-		parent::__construct($backend, $originalLocation, $deletedTime, $trashPath, $fileInfo, $user);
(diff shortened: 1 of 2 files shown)


artonge:
> Ideally we should store the information of who deleted the item and pass it, but for now just pass null to avoid a crash.

I think @Pytal added a DB table for that. But it might not be support by groupfolders yet

come-nc:
> > Ideally we should store the information of who deleted the item and pass it, but for now just pass null to avoid a crash.
> 
> I think @Pytal added a DB table for that. But it might not be support by groupfolders yet

It’s a column in oc_files_trash: https://github.com/nextcloud/server/pull/44643/files#diff-ef08ec57acd193f58ac8dd372a6c4055ea7fe925243ebd9a02be9a6fff209d3bR48
groupfolders trash backend has its own table so it needs to be modified to do the same thing.

Pytal (approved):

Issue #2822, opened 2024-02-14
Prevent self-lockout with advanced permissions


ctueck:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

Our common pattern for group folders is: a wider group (all staff) has full access to the folder by default. Another group (managers, subset of staff) has the right to manage advanced permissions and occasionally wants to create sub-folders with full rights for managers (or certain other users or subgroups) only, and limited or no rights for the wider staff group.

When the intention is to give no rights (i.e. not even read) to the wider group, it all works well as long as the manager first adds an advanced permission rule to give themselves full access, and then creates a rule to deny access for the wider group. But...

We have had a few instances of managers locking everyone, including themselves, out of a sub-folder by first creating a deny-all rule for the wider staff group.

I understand it's logical that it happens, but the current workings make this mistake quite easy to happen. And as the folder becomes entirely invisible, this can only be fixed via `occ` commands.

**Describe the solution you'd like**

Would it be reasonable to consider users who have the right to manage advanced permissions as kind of super-users for that group folder and always entitle them to full access regardless of any advanced permission settings?

**Describe alternatives you've considered**

Two other alternatives that came to my mind:

1.  If a user has no read permissions for a file or folder, but read permission for the containing folder, they can still see the file/folder but not enter/open/download it. (And should be able to manage advanced permissions provided they have that privilege.) This would not prevent the mistake, but make it easy and straight-forward for regular users to recover from it. Also, this would be more in line with regular Unix-like permissions logic - maybe this would actually be the better alternative.

2. Implement some kind of warning before someone creates a rule that would make the file disappear for them (or everyone). But this might be unnecessarily complex. It's also questionable whether it should be possible (via the GUI) to create a rule that makes a file/folder inaccessible and hidden from everyone to such an extent that even oneself cannot get it back, at least not via the GUI/without resorting to an `occ` command.

**Additional context**

Running NC 27.1.5 and groupfolders 15.3.4

I tried to check for any duplicate/closely related issue, but didn't find any and hope I didn't miss something.

github-cli:
just a user reading this...

[...]

PR #2960, opened 2024-05-14
fix(ACL): Add check to prevent users from revoking their own access

icewind1991:
When changing ACL rules, check if the new rule set would lead the current user without read permissions

- [x] make the UI properly handle the error
- [ ] ~(maybe) add an option to force revoking the access anyway~

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
 
 namespace OCA\GroupFolders\ACL;
 
+use OCA\GroupFolders\ACL\UserMapping\IUserMappingManager;
 use OCA\GroupFolders\Trash\TrashManager;
 use OCP\Cache\CappedMemoryCache;
 use OCP\Constants;
@@ -22,6 +23,7 @@ class ACLManager {
 	public function __construct(
 		private readonly RuleManager $ruleManager,
 		private readonly TrashManager $trashManager,
+		private readonly IUserMappingManager $userMappingManager,
 		private readonly LoggerInterface $logger,
 		private readonly IUser $user,
 		private readonly \Closure $rootFolderProvider,
@@ -85,7 +87,7 @@ class ACLManager {
 		$fromTrashbin = str_starts_with($path, '__groupfolders/trash/');
 		if ($fromTrashbin) {
 			/* Exploded path will look like ["__groupfolders", "trash", "1", "folderName.d2345678", "rest/of/the/path.txt"] */
-			[,,$groupFolderId,$rootTrashedItemName] = explode('/', $path, 5);
+			[, , $groupFolderId, $rootTrashedItemName] = explode('/', $path, 5);
 			$groupFolderId = (int)$groupFolderId;
 			/* Remove the date part */
 			$separatorPos = strrpos($rootTrashedItemName, '.d');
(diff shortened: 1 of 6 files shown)


jancborchardt (commented):
@icewind1991 since you said

> Would it maybe make more sense to just fully prevent users from revoking their own access?

I would say the modal is better, because maybe you do want to get out of the group folder?

> **Access removal**
> Proceeding will remove your own access to this group folder.
> [ Cancel ]  [ Remove own access ]

provokateurin:
Fixes https://github.com/nextcloud/groupfolders/issues/2822

provokateurin:
> I would say the modal is better, because maybe you do want to get out of the group folder?

I don't think this makes sense, since in fact you do not remove yourself from the Team folder this way.
If the user no longer wants access, then an admin has to remove their access and they can not do it themselves through ACL (which also only works on sub folders, not the root Team folder itself anyway).

come-nc (approved):

Jerome-Herbinet:
Hi, awesome improvement ! Can't it be backported to some previous major releases @icewind1991 @jancborchardt ?

provokateurin:
Let's try!

Jerome-Herbinet:
> @Jerome-Herbinet stable30 was relatively easy to do, but stable29 unfortunately would need a larger amount of work, so not backporting it there.

Thanks for your answer @provokateurin. Can you tell me when the next version of Team Folders for Nextcloud 30 (> 18.1.0) will be released (the version that will include this new backported feature)?

provokateurin:
I can do it next Monday, should be 18.1.1 indeed.

Jerome-Herbinet:
> I can do it next Monday, should be 18.1.1 indeed.

It would be great @provokateurin ! I'm currently making (and scheduling) a communication talking about this for next Tuesday (on Arawa's Mastodon, BlueSky and LinkedIn pages).

provokateurin:
@Jerome-Herbinet I just published 18.1.1 already :)

Jerome-Herbinet:
> @Jerome-Herbinet I just published 18.1.1 already :)

@provokateurin thanks for your work ! One small issue by the way (tell me if you want me to open an issue) : 

Here is the error message on NC 31 : 
Capture d’écran du 2025-03-06 08-41-38

Here is the error message on NC 30  : 
Capture d’écran du 2025-03-06 08-40-22

Is the NC 30 error message what you expected?

provokateurin:
Hm this is what I saw on 29, but not on 30 when I tested it. On 29 I looked into it and realized there was quite a bit of error handling missing, while on 30 it was fine.

How did you install the app? If you built it yourself please make sure the frontend is recompiled before trying again.

Jerome-Herbinet:
> Hm this is what I saw on 29, but not on 30 when I tested it. On 29 I looked into it and realized there was quite a bit of error handling missing, while on 30 it was fine.
> 
> How did you install the app? If you built it yourself please make sure the frontend is recompiled before trying again.

I made a simple app upgrade via the web UI @provokateurin

Jerome-Herbinet:
@provokateurin 

I upgraded the app on another NC 31 test instance  (with same regular upgrade process via web UI). Same error message : 

Capture d’écran du 2025-03-06 09-57-47

provokateurin:
Can you open an issue about it? I won't be able to fix it before your presentation, but I hope it won't block you!

Jerome-Herbinet:
> Can you open an issue about it? I won't be able to fix it before your presentation, but I hope it won't block you!

No problem, I'll do it this morning (and I will adapt or delay my communication) 🙂 Thanks !

Jerome-Herbinet:
Done @provokateurin : https://github.com/nextcloud/groupfolders/issues/3670

PR #2976, opened 2024-05-29
Fix TypeError on versions expire

solracsf:


lib/Versions/GroupVersion.php
jonaszkadziela, line 1: The return type of the `getVersionFile` method should also be changed to `FileInfo`.

Otherwise, I'm getting the following error:
`TypeError: OCA\GroupFolders\Versions\GroupVersion::getVersionFile(): Return value must be of type OCP\Files\File, OC\Files\FileInfo returned in /var/www/html/custom_apps/groupfolders/lib/Versions/GroupVersion.php:51`

solracsf: Hum, maybe; but then I'm expecting it will fail here 🤔 
https://github.com/nextcloud/groupfolders/blob/13a68f8d59d268687274006f74f947836dc63213/lib/Versions/VersionsBackend.php#L238

Let's see what CI thinks.

solracsf: @jonaszkadziela check https://github.com/nextcloud/groupfolders/actions/runs/9336320832/job/25696721210?pr=2976#step:6:18

jonaszkadziela: @solracsf Indeed, this change results in a pipeline failure 😞

I'm not that familiar with Nextcloud codebase, but from what I understand, the `File` interface extends the `Node` interface, which extends the `FileInfo` interface.

After your recent commits, I'm still a TypeError:
image

Inside the `VersionsBackend` class, we are creating a new `GroupVersion` object and passing a variable of `FileInfo` type. This will cause the exception because polymorphism will not work backwards (`File` is an instance of `FileInfo`, but `FileInfo` is not an instance of `File`).

I think in that case, the `getVersionsForFileFromDB` method needs to be changed, to accept the parameter of type `File` instead of `FileInfo`. Another option is to rearrange the code in the `GroupVersion` class so that it works with the `FileInfo` type

Issue #2952, opened 2024-05-10
Undefined array key "group_details"


RenatoMorenoR:
Hello friends, we have a problem in our platform, we see a constant error that appeared a couple of weeks ago.

Every time we want to add a group directory via API we get the error

“method”: “POST”, “url”:“/index.php/apps/groupfolders/folders”, “message”: “Undefined array key \”group_details“ at /usr/share/webapps/nextcloud/apps/groupfolders/lib/Controller/FolderController.php#307”, “userAgent”:“--”, “version”: “28.0.3.2”, “data”:{“app”: “PHP”}

### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. add a group directory from the API

### Expected behavior
You should be able to create the directory without error in the error logo.

### Actual behavior
The directory is created without problem, but sends the error message to the system.

This is the problem function
```
private function folderDataForXML(array $data): array {
		$groups = $data['group_details'];
		$data['groups'] = [];
		unset($data['group_details']);
		foreach ($groups as $id => $group) {
			$data['groups'][] = [
				'@group_id' => $id,
				'@permissions' => $group['permissions'],
				'@display-name' => $group['displayName'],
				'@type' => $group['type'],
			];
		}
		return $data;
	}
```
Perhaps if you include a simple validation as shown here, you could avoid this error. 
```
private function folderDataForXML(array $data): array {
                if(!is_null($data['group_details'])) {
		  $groups = $data['group_details'];
		  $data['groups'] = [];
		  unset($data['group_details']);
		  foreach ($groups as $id => $group) {
			$data['groups'][] = [
				'@group_id' => $id,
				'@permissions' => $group['permissions'],
				'@display-name' => $group['displayName'],
				'@type' => $group['type'],
			];
		  }
                }
		return $data;
	}
```

### Server configuration

**Operating system:**
Ubuntu 22.04

**Web server:**
Nginx

**Database:**
Mysql
**PHP version:**
8.1
**Nextcloud version:** (see Nextcloud admin page)
28.0.3.2
**Group folders version:**
16.0.6
**Updated from an older Nextcloud/ownCloud or fresh install:**
Yes
[...]

PR #3042, opened 2024-07-05
fix: check if array key exists

solracsf:
Fix https://github.com/nextcloud/groupfolders/issues/2952

### DIFF
--- a/lib/Controller/FolderController.php
+++ b/lib/Controller/FolderController.php
@@ -304,9 +304,14 @@ class FolderController extends OCSController {
 	}
 
 	private function folderDataForXML(array $data): array {
-		$groups = $data['group_details'];
+		$groups = [];
+
+		if (isset($data['group_details'])) {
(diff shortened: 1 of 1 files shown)


lib/Controller/FolderController.php
```diff
@@ -304,9 +304,14 @@ private function buildOCSResponseXML(string $format, DataResponse $data): V1Resp
...
 
 	private function folderDataForXML(array $data): array {
-		$groups = $data['group_details'];
+		$groups = [];
+
+		if (isset($data['group_details'])) {
+			$groups = $data['group_details'];
+			unset($data['group_details']);
+		}
+
```
come-nc, line 313:
```suggestion
		$groups = $data['group_details'] ?? [];
		unset($data['group_details']);
```
This should work too and is shorter.

come-nc (approved):
Small improvement suggestion but otherwise good

artonge (approved):

PR #3067, opened 2024-07-18
feat(ACL): Show actual inherited permissions

Fenn-CS:
### Photos

| Before | After |
| ---------- | ------- |
| inherited-masked|inherited-real |

### TODO
- [x] Increase contrast for inherited icons as is now
- [x] Remove the dash completely as "inherited" does not necessarily mean blank?
- [ ] Better screenshot of after having but allowed and not allowed?


provokateurin:
https://github.com/nextcloud/groupfolders/pull/4126

PR #3098, opened 2024-08-07
fix: folder icon

skjnldsv:
image

### Before
image


AndyScherzinger (approved):
LGTM but lacks SVG height/width which is an issue for Safari browsers from wehat I understood from nickvergessen or at least some iOS devices (and their Safari)

skjnldsv:
> LGTM but lacks SVG height/width which is an issue for Safari browsers from wehat I understood from nickvergessen or at least some iOS devices (and their Safari)

Adjusted! :+1: 
Thanks :bow:

jancborchardt (approved):
Thanks, nice! :)

PR #3160, opened 2024-09-03
feat(settings): Make default quota configurable

provokateurin:


lib/Controller/FolderController.php
```diff
@@ -145,10 +146,15 @@ private function getRootFolderStorageId(): ?int {
...
+	 * @throws OCSNotFoundException
 	 */
 	public function addFolder(string $mountpoint): DataResponse {
 		$id = $this->manager->createFolder(trim($mountpoint));
-		return new DataResponse(['id' => $id]);
+		$folder = $this->manager->getFolder($id, $this->rootFolder->getMountPoint()->getNumericStorageId());
+		if ($folder === false) {
+			throw new OCSNotFoundException();
+		}
+		return new DataResponse($folder);
```
artonge, line 157: This changes the response type, is `$folder` containing `'id'`?

README.md
```diff
@@ -50,6 +50,12 @@ For entitlements, only users from those groups are selectable which have to be c
 
 ![advanced permission entitlement](screenshots/aclAdmin.png)
 
+## Configuration parameters
+
+Some settings are currently only exposed via `config/config.php`:
+
+Configures the default quota used when creating new groupfolders: `'groupfolders.quota.default' => -3,`
```
artonge, line 57: Maybe add the meaning of special values like `-3`.

lib/Migration/Version19000Date20240903062631.php
```diff
@@ -0,0 +1,42 @@
...
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('group_folders')) {
+			$table = $schema->getTable('group_folders');
+			$table->changeColumn('quota', [
+				'notnull' => true,
+				'length' => 6,
+				'default' => null,
```
artonge, line 36: `notnull === true` and `default === null`. Is it not contradictory?

lib/Controller/FolderController.php
provokateurin: Yes

README.md
provokateurin: I was searching for some general documentation about the meaning, but I couldn't find any. Do you know if we have something like that?

lib/Migration/Version19000Date20240903062631.php
provokateurin: No, it just means you need to provide a value when inserting a new row. Previously you could omit it because of the default value.

README.md
artonge: From the code, it looks like the only special value is `-3`. You can also specify the unit of the value, which seem to be in bits.

provokateurin: It's bytes ;)

artonge (approved):

sorbaugh (approved):

PR #3161, opened 2024-09-03
fix(settings): Calculate correct default quota option sizes

provokateurin:
I first noticed the unit was wrong, decimal vs binary notation.
Then I realized the values were wrong too, not sure why this is so messed up:
```
1073741274 -1024^3   =-550
5368709120 -1024^3*5 = 0
10737412742-1024^3*10=-5498
```
Really weird how one is correct, one is off by a bit and the other that should be 10x off is slight less off than 10x??

Should we add a migration that corrects these quota values if they are configured for existing groupfolders?

### DIFF
--- a/src/settings/App.tsx
+++ b/src/settings/App.tsx
@@ -18,10 +18,11 @@ import AdminGroupSelect from './AdminGroupSelect'
 import SubAdminGroupSelect from './SubAdminGroupSelect'
 import { loadState } from '@nextcloud/initial-state'
 
+const bytesInOneGibibyte = Math.pow(1024, 3)
 const defaultQuotaOptions = {
-	'1 GB': 1073741274,
-	'5 GB': 5368709120,
-	'10 GB': 10737412742,
+	'1 GB': bytesInOneGibibyte,
+	'5 GB': bytesInOneGibibyte * 5,
+	'10 GB': bytesInOneGibibyte * 10,
 	Unlimited: -3,
 }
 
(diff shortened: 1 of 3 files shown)


artonge (approved):
A migration sounds good, else admins will see a custom set quota, no?

provokateurin:
Yeah, it looks weird now as it says 1,024 GiB for the previous 1 GB option.

provokateurin:
Repair step added

src/settings/App.tsx
```diff
@@ -18,10 +18,11 @@ import AdminGroupSelect from './AdminGroupSelect'
...
 import { loadState } from '@nextcloud/initial-state'
 
+const bytesInOneGibibyte = Math.pow(1024, 3)
 const defaultQuotaOptions = {
-	'1 GB': 1073741274,
-	'5 GB': 5368709120,
-	'10 GB': 10737412742,
+	'1 GB': bytesInOneGibibyte,
+	'5 GB': bytesInOneGibibyte * 5,
+	'10 GB': bytesInOneGibibyte * 10,
```
Altahrim, line 25: Should we change the unit for clarity?
```suggestion
	'1 GiB': bytesInOneGibibyte,
	'5 GiB': bytesInOneGibibyte * 5,
	'10 GiB': bytesInOneGibibyte * 10,
```

provokateurin: I did that in the beginning, but then I figured out this is the style that is used everywhere.
I don't like it either as it is the worst of both worlds, but we'll just have to live with it.

provokateurin:
Can someone merge? The failures are unrelated to the changes :/

PR #3183, opened 2024-09-09
fix(versions): Handle missing versions folder

provokateurin:
The catch references an potentially uninitialized variable and thus will fail if executed.

### DIFF
--- a/lib/Versions/VersionsBackend.php
+++ b/lib/Versions/VersionsBackend.php
@@ -67,11 +67,12 @@ class VersionsBackend implements IVersionBackend, IMetadataVersionBackend, IDele
 
 	public function getVersionFolderForFile(FileInfo $file): Folder {
 		$folderId = $this->getFolderIdForFile($file);
+		$groupfoldersVersionsFolder = $this->getVersionsFolder($folderId);
 
 		try {
-			$groupfoldersVersionsFolder = $this->getVersionsFolder($folderId);
 			/** @var Folder $versionsFolder */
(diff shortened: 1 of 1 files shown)


lib/Versions/VersionsBackend.php
```diff
@@ -67,11 +67,12 @@ private function getFolderIdForFile(FileInfo $file): int {
...
 	public function getVersionFolderForFile(FileInfo $file): Folder {
 		$folderId = $this->getFolderIdForFile($file);
+		$groupfoldersVersionsFolder = $this->getVersionsFolder($folderId);
 
 		try {
-			$groupfoldersVersionsFolder = $this->getVersionsFolder($folderId);
 			/** @var Folder $versionsFolder */
-			return $groupfoldersVersionsFolder->get((string)$file->getId());
+			$versionsFolder = $groupfoldersVersionsFolder->get((string)$file->getId());
+			return $versionsFolder;
```
artonge, line 75:
```suggestion
			return $groupfoldersVersionsFolder->get((string)$file->getId());
```

artonge (approved):

lib/Versions/VersionsBackend.php
provokateurin: This is necessary to assign the type because it returns Node but we know it is Folder and the method expects only Folder to be returned. If not done here it would need to be done in https://github.com/nextcloud/groupfolders/pull/3184 anyway.

provokateurin:
The method didn't exist back in stable28 yet, so backport failure is not a problem.

PR #3192, opened 2024-09-10
fix(ACL): Actually return in ACLStorageWrapper methods

provokateurin:
Psalm caught it but it was just ignored through the baseline :disappointed:


artonge (approved):
Oo

PR #3201, opened 2024-09-10
refactor(settings): Remove usage of jQuery

provokateurin:


src/settings/Api.ts
```diff
@@ -81,79 +94,72 @@ export class Api {
 		return axios.post(generateUrl('/apps/settings/') + '/settings/authorizedgroups/saveSettings', {
 			newGroups,
 			class: classname,
-		}).then((data) => data.data)
```
artonge, line 84: Don't we need to keep it? `response.data` ?

artonge (approved):

src/settings/Api.ts
provokateurin: The return type is Promise<void> and the result is never used anywhere, so I removed it.

artonge: Ah, then we should probably remove the return statement

artonge: Or change the return type

provokateurin: Returning the promise is fine otherwise it would be await. There is really no difference in that.

provokateurin:
Cypress failure is due to https://github.com/nextcloud/server/pull/47848#issuecomment-2343132766. @artonge can you merge?

src/settings/Api.ts
```diff
@@ -3,9 +3,23 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 import { generateUrl, generateOcsUrl } from '@nextcloud/router'
-import { OCSResult, AxiosOCSResult } from 'NC'
 import axios from '@nextcloud/axios'
 
+interface AxiosOCSResult<T> {
```
susnux, line 8: No need for custom interfaces, you could also just use the interface from `@nextcloud/typings`.

src/settings/Api.ts
```diff
@@ -81,79 +94,72 @@ export class Api {
...
-		return $.post(this.getUrl(`folders/${folderId}/mountpoint`), {
+		return axios.post(this.getUrl(`folders/${folderId}/mountpoint`), {
 			mountpoint
 		})
 	}
 
 	async aclMappingSearch(folderId: number, search: string): Promise<{groups: ManageRuleProps[], users: ManageRuleProps[]}> {
-		return $.getJSON(this.getUrl(`folders/${folderId}/search?format=json&search=${search}`))
-			.then((data: OCSResult<{ groups: OCSGroup[]; users: OCSUser[]; }>) => {
+		return axios.get(this.getUrl(`folders/${folderId}/search`), { params: { search } })
```
susnux, line 152: Why is this an async function but then no await is used but `.then()` chain?
(I know this is unrelated to your changes but no good code style)

src/settings/Api.ts
```diff
@@ -81,79 +94,72 @@ export class Api {
...
 	async setACL(folderId: number, acl: boolean): Promise<void> {
-		return $.post(this.getUrl(`folders/${folderId}/acl`), {
+		return axios.post(this.getUrl(`folders/${folderId}/acl`), {
 			acl: acl ? 1 : 0
 		})
 	}
 
 	async renameFolder(folderId: number, mountpoint: string): Promise<void> {
-		return $.post(this.getUrl(`folders/${folderId}/mountpoint`), {
+		return axios.post(this.getUrl(`folders/${folderId}/mountpoint`), {
```
susnux, line 146: Similar here, an async function without an await. Pretty bad code style.
Also it returns in a `void` function.

```suggestion
		await axios.post(this.getUrl(`folders/${folderId}/mountpoint`), {
```

susnux (approved):
Nice refactoring, some comments but not about your code :sweat_smile:

src/settings/Api.ts
provokateurin: Right, it could be sync and still return the Promise as it already does.

src/settings/Api.ts
provokateurin: Ah I didn't know this was available, I only moved the interface into the file.

src/settings/Api.ts
provokateurin: It's a Promise of void which is not the same as void.

PR #3205, opened 2024-09-10
Fix slow move on same object bucket during file upload.

cfiehe:
This commit fixes the issue https://github.com/nextcloud/server/issues/47856. When you upload a file into a group folder and when you use a single S3 bucket as primary storage, the final move operation hangs for a long time. In the background, Nextcloud initiates a copy-delete sequence from the bucket into the bucket, with causes a lot unnecessary overhead. Nextcloud thinks that the file must be imported to another storage and does not recognize that everything is done on the same object bucket. In that case, the import step can be completely skipped, which saves time, network bandwidth and reduces the load on the object storage.

### DIFF
--- a/lib/Mount/GroupFolderStorage.php
+++ b/lib/Mount/GroupFolderStorage.php
@@ -11,6 +11,7 @@ use OC\Files\ObjectStore\ObjectStoreScanner;
 use OC\Files\ObjectStore\ObjectStoreStorage;
 use OC\Files\Storage\Wrapper\Quota;
 use OCP\Files\Cache\ICacheEntry;
+use OCP\Files\Storage\IStorage;
 use OCP\IUser;
 use OCP\IUserSession;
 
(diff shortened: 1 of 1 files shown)


lib/Mount/GroupFolderStorage.php
```diff
@@ -69,4 +69,14 @@ public function getScanner($path = '', $storage = null) {
 		}
 		return $storage->scanner;
 	}
+
+	public function moveFromStorage(IStorage $sourceStorage, $sourceInternalPath, $targetInternalPath) {
+		if ($sourceStorage->instanceOfStorage(ObjectStoreStorage::class) &&
+			$this->instanceOfStorage(ObjectStoreStorage::class) &&
+			$sourceStorage->getObjectStore()->getStorageId() == $this->getObjectStore()->getStorageId()) {
```
susnux, line 76:
```suggestion
			$sourceStorage->getObjectStore()->getStorageId() === $this->getObjectStore()->getStorageId()) {
```

artonge (commented):
Looks good, but would wait for @icewind1991 opinion. Maybe this logic should be moved to the server repo.

provokateurin:
> Maybe this logic should be moved to the server repo.

I'm  pretty sure this is the case. It's also something I suggested in https://github.com/nextcloud/server/issues/47682 as I noticed the same problem when you only upload a single chunk.

cfiehe:
Well, I think the fix can stay in the plugin until Nextcloud has developed further and handles file renaming on the same storage in a more efficient way than today. Unfortunately, I cannot find something similar in case of a local storage. Uploads in a group folder on a local storage always cause an inefficient copy-delete on Nextcloud triggered by the `ChunkingPlugin` and the general fallback in `OCA\DAV\Connector\Sabre\Directory`. Currently, there seems no way to get there in between.

sorbaugh:
Nice PR! Let's wait indeed for @icewind1991. If everything is ok, I'd say we merge this one as it would be a significant performance boost. We can then STILL move the logic to server :)

icewind1991:
This logic should probably go into `ObjectStoreStorage` instead to handle more cases around "cross-storage moves" that stay within the same object store.

Also instead of doing nothing/`return true` it should still trigger the `moveFromCache` to ensure that cache stays up to date.

cfiehe:
@icewind1991 @provokateurin
I have created a new pull request https://github.com/nextcloud/server/pull/47986. May someone please have a close look on it. I hope, that the change does not have any negative side effects.

provokateurin:
Closing this in favor of the server PR :)

Issue #3095, opened 2024-08-05
data loss: named versions are lost from group folder after a while


sena73:
"Versions" extension is not working properly with the "Group folder" extension in NC 29. 

If I create named versions of a file  in a group folder, they are created, but disappear after a while (noticed after days). 

Old versions simply disappear in a group folder without any notice or log message. I would tell it is a severe bug, as I do not see a way to recover deleted versions.

I understand that it gonna take time to fix.  But please, make at least this 2 extensions conflicting, so people will not be able to install both of them simultaneously. Or better make "Versions" menu disabled for the files in "Group folder".

I found other bugs that indicate problems when both extensions are working: 

#2965 #2343

joshtrichards:
> If I create named versions of a file in a group folder, they are created, but disappear after a while (noticed after days).

You may be correct. I suspect we're expiring labeled (named) versions within GF. Offhand I don't see the GF Versions handler checking for labeled versions in the same way that the non-GF versions does:
 
https://github.com/nextcloud/server/blob/2480567005776cc96d543b52992c83a41d75d1a6/apps/files_versions/lib/Storage.php#L569-L570

This may have been lost during the #2543 -> #2857 jump though offhand I don't see it in the former either.

Cc: @artonge

artonge:
@joshtrichards is right, we do have the check for "normal" versions, but I missed to implement it for groupfolders' versions.

joshtrichards:
Thanks for reporting this, @sena73. Will be fixed in the next release!

XueSheng-GIT:
@artonge @joshtrichards I've upgraded groupfolders to 17.0.4 (NC 28.0.7) which includes the backport of https://github.com/nextcloud/groupfolders/pull/3213 but named versions still expire!

Can you double check and re-open this issue? Or shall I create a new one?

joshtrichards:
I believe that API is only applicable to >28. I think we'd have to do something unique just for 28 🤔. I'm not certain offhand.

Are you sure you're running 17.0.4 though? It is only for >=29.

XueSheng-GIT:
@joshtrichards I'm on NC 29.0.7 (28 was a typo). Sorry for that confusion.

XueSheng-GIT:
Just in case somebody stumbles over this, I now changed the following as quick fix to stop auto expiry of labelled versions. Maybe not the cleanest solution but it seems to work.

```
[...]

PR #3213, opened 2024-09-13
fix: Do not expire labeled versions

artonge:
Fix: https://github.com/nextcloud/groupfolders/issues/3095

### DIFF
--- a/lib/Versions/ExpireManager.php
+++ b/lib/Versions/ExpireManager.php
@@ -9,6 +9,7 @@ declare(strict_types=1);
 namespace OCA\GroupFolders\Versions;
 
 use OCA\Files_Versions\Expiration;
+use OCA\Files_Versions\Versions\IMetadataVersion;
 use OCA\Files_Versions\Versions\IVersion;
 
 /**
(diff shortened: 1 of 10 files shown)


lib/Versions/ExpireManager.php
```diff
@@ -10,6 +10,7 @@
 
 use OCA\Files_Versions\Expiration;
 use OCA\Files_Versions\Versions\IVersion;
+use OCA\Files_Versions\Versions\IMetadataVersion;
```
come-nc, line 13: You need to add a stub for it

lib/Versions/ExpireManager.php
```diff
@@ -117,6 +118,16 @@ public function getExpiredVersion(array $versions, int $time, bool $quotaExceede
 		});
 
 		$expired = array_filter($versionsLeft, function (IVersion $version) use ($quotaExceeded) {
+			// Do not expire current version.
+			if ($version instanceof IMetadataVersion && $version->getTimestamp() === $version->getSourceFile()->getMtime()) {
+				return false;
+			}
+
+			// Do not expire versions with a label.
+			if ($version->getMetadataValue('label') !== null && $version->getMetadataValue('label') !== '') {
```
come-nc, line 127: It’s a bit weird/sad that there is not a method like isExpirable in versions instead. It seems a bit arbitrary to decide from groupfolders app what the rules are for keeping versions.

come-nc (approved):
Okay apart from missing stub.

PR #3225, opened 2024-09-15
fix(psalm): Add all missing stubs

provokateurin:
Generated using https://github.com/provokateurin/php-stubs-updater with this command: `update-stubs.php tests/stubs ../../3rdparty/doctrine/dbal/ ../../3rdparty/icewind/streams/ ../../3rdparty/symfony/console/ ../../3rdparty/stecman/symfony-console-completion/ ../../lib/ ../../tests/lib/ ../../core/ ../dav/lib/ ../settings/lib/ ../circles/lib/ ../files/lib/ ../files_versions/lib/ ../files_trashbin/lib/ ../files_sharing/lib/ ../richdocuments/lib`

Not sure if we should backport this, all the bugs it found do not lead to any runtime errors.


lib/ACL/ACLCacheWrapper.php
```diff
@@ -64,19 +64,19 @@ public function getFolderContentsById($fileId) {
...
 	}
 
 	public function searchByMime($mimetype) {
 		$results = $this->getCache()->searchByMime($mimetype);
 		$this->preloadEntries($results);
-		return array_map([$this, 'formatCacheEntry'], $results);
+		return array_filter(array_map([$this, 'formatCacheEntry'], $results));
 	}
 
 	public function searchQuery(ISearchQuery $query) {
```
come-nc, line 76:
```suggestion
	public function searchQuery(ISearchQuery $searchQuery) {
```

lib/AppInfo/Application.php
```diff
@@ -228,7 +229,7 @@ public function register(IRegistrationContext $context): void {
 	}
 
 	public function boot(IBootContext $context): void {
-		$context->injectFn(function (IMountProviderCollection $mountProviderCollection, CacheListener $cacheListener, IGroupManager $groupManager): void {
+		$context->injectFn(function (IMountProviderCollection $mountProviderCollection, CacheListener $cacheListener, Group\Manager $groupManager): void {
```
come-nc, line 232: Is it possible to use IGroupManager&Emitter or something?
Or even drop the listen calls and listen to the events instead. Pretty sure we have typed events for groups.

lib/DAV/GroupFoldersHome.php
```diff
@@ -77,7 +77,7 @@ private function getFolder(string $name): ?array {
 	private function getDirectoryForFolder(array $folder): GroupFolderNode {
 		$userHome = '/' . $this->user->getUID() . '/files';
 		$node = $this->rootFolder->get($userHome . '/' . $folder['mount_point']);
-		return new GroupFolderNode(Filesystem::getView($userHome), $node, $folder['folder_id']);
+		return new GroupFolderNode(Filesystem::getView(), $node, $folder['folder_id']);
```
come-nc, line 80: Hum ugly, looks like original intent was something like new View($userHome), but changing this now will have effects, should double check what should be done.
If $userHome is not used anymore it shoudl also be removed.

lib/Folder/FolderManager.php
```diff
@@ -482,7 +482,7 @@ public function searchUsers(int $id, string $search = '', int $limit = 10, int $
 	/**
 	 * @param string $groupId
 	 * @param int $rootStorageId
-	 * @return list<array{folder_id: int, mount_point: string, permissions: int, quota: int, acl: bool, rootCacheEntry: ?ICacheEntry}>
+	 * @return list<array{folder_id: int, mount_point: string, permissions: int, quota: int, acl: bool, rootCacheEntry: ?CacheEntry}>
```
come-nc, line 485: Why is this change needed? Any way to avoid the OC ref?

lib/Mount/GroupFolderNoEncryptionStorage.php
```diff
@@ -10,5 +10,6 @@
 
 use OCP\Files\Storage\IDisableEncryptionStorage;
 
+/** @psalm-suppress UnimplementedInterfaceMethod */
```
come-nc, line 13: Add a comment explaining why you suppress.

come-nc (changes requested):

lib/Folder/FolderManager.php
provokateurin: Not really, as we use a private class it specifies the private type is returned. So only way to really fix it is by only using public classes.

lib/DAV/GroupFoldersHome.php
provokateurin: The logic of getView() already returns a View for the user home by default if a user exists, so nothing changes. $userHome is still used to construct the $node so that's why it is not removed.

lib/AppInfo/Application.php
provokateurin: Ah I tried it, but our DI gets confused and it does not actually work. So private class it is then.

provokateurin:
Sorry for the mess :sweat:

provokateurin:
And one more to make reuse happy :/

icewind1991:
the 3rdparty libs should probably be added as dev-dependency instead of adding them as stubs

provokateurin:
To get updates automatically? Or what would be the reason?

provokateurin:
If we add them as dev dependencies they don't necessary follow the same constraints as in 3rdparty, so I would rather have the stubs where we know that they exactly match the currently used versions.

icewind1991 (approved):

PR #3234, opened 2024-09-16
fix: Enforce strict types everywhere

provokateurin:
Depends on https://github.com/nextcloud/server/pull/48094 as View.php can't handle false for getOwner() everywhere.
All untouched files already had it enabled.


provokateurin:
Interesting, in CI it doesn't fail without the PR mentioned above.

provokateurin:
Code coverage in groupfolders is only around 30% which is way to little to turn this on with confidence.

PR #3252, opened 2024-09-19
Refactor with rector

provokateurin:


rector.php
```diff
@@ -15,4 +15,10 @@
 	->withSkip([
 		__DIR__ . '/tests/stubs',
 	])
-	->withTypeCoverageLevel(0);
+	->withPreparedSets(
+		deadCode: true,
+		typeDeclarations: true,
+		earlyReturn: true,
```
come-nc, line 21: This decrease readability for no benefit, please revert the commit

rector.php
```diff
@@ -15,4 +15,10 @@
 	->withSkip([
 		__DIR__ . '/tests/stubs',
 	])
-	->withTypeCoverageLevel(0);
+	->withPreparedSets(
+		deadCode: true,
+		typeDeclarations: true,
```
come-nc, line 20: What is the benefit of this over withTypeCoverageLevel? Is it like maximum level?

come-nc (changes requested):

rector.php
provokateurin: Yes

come-nc (approved):

PR #3257, opened 2024-09-19
fix: Add missing null checks

provokateurin:


lib/Command/FolderCommand.php
```diff
@@ -40,7 +40,14 @@ protected function getFolder(InputInterface $input, OutputInterface $output): ?a
 			return null;
 		}
 
-		$folder = $this->folderManager->getFolder($folderId, $this->rootFolder->getMountPoint()->getNumericStorageId());
+		$rootStorageId = $this->rootFolder->getMountPoint()->getNumericStorageId();
+		if ($rootStorageId === null) {
+			$output->writeln('<error>storage id not found ' . $rootStorageId . '</error>');
```
come-nc, line 45: You are concatenating with null here, not really useful.
```suggestion
			$output->writeln('<error>Root storage id not found</error>');
```

lib/Command/ListCommand.php
```diff
@@ -55,6 +55,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		}
 
 		$rootStorageId = $this->rootFolder->getMountPoint()->getNumericStorageId();
+		if ($rootStorageId === null) {
+			$output->writeln('<error>storage id not found ' . $rootStorageId . '</error>');
```
come-nc, line 59: same

lib/Controller/FolderController.php
```diff
@@ -73,7 +78,12 @@ private function formatFolder(array $folder): array {
 	#[NoAdminRequired]
 	#[ApiRoute(verb: 'GET', url: '/folders')]
 	public function getFolders(bool $applicable = false): DataResponse {
-		$folders = $this->manager->getAllFoldersWithSize($this->getRootFolderStorageId());
+		$storageId = $this->getRootFolderStorageId();
+		if ($storageId === null) {
+			throw new OCSNotFoundException();
```
come-nc, line 83: Should be documented no? Isn’t the openapi extractor checking that?

provokateurin: Yeah, but I will use a psalm rule for that to enforce documenting all possible exceptions. So expect a follow up for that.

come-nc (approved):

PR #3261, opened 2024-09-20
test: Add tests for Events

provokateurin:


come-nc (commented):
Are those kind of tests usefull? They really feel like tautologies.

provokateurin:
I think it's good to have them so that we can easily adapt the tests if the events ever get some more logic that would need tests.

come-nc (approved):
:shrug:

PR #3270, opened 2024-09-21
fix(psalm): Fix all @throws annotations and add missing ones

provokateurin:


lib/AppInfo/Application.php
```diff
@@ -227,6 +229,10 @@ public function register(IRegistrationContext $context): void {
 		$context->registerMiddleware(AuthorizedAdminSettingMiddleware::class);
 	}
 
+	/**
+	 * @throws ContainerExceptionInterface
+	 * @throws Throwable
```
come-nc, line 234: When does that throw Throwable? For me that should never be documented, Throwable which are not Exception are bugs to be fixed, not documented. Anything can throw a Throwable if you use it wrong.
Would also be good if those `@throws` tag would include a description of when to expect a throw.

lib/AppInfo/Capabilities.php
```diff
@@ -21,6 +23,10 @@ public function __construct(
 	) {
 	}
 
+	/**
+	 * @throws Exception
+	 * @throws RequestBuilderException
+	 */
```
come-nc, line 29: Should be caught inside, this method is not documented as throwing in ICapability and should not throw.

lib/Command/Create.php
```diff
@@ -10,17 +10,26 @@
...
+use OCP\DB\Exception;
+use Symfony\Component\Console\Exception\InvalidArgumentException;
+use Symfony\Component\Console\Exception\LogicException;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class Create extends Base {
+	/**
+	 * @throws LogicException
```
come-nc, line 22: when? Add a description please

lib/Command/ExpireGroup/ExpireGroupBase.php
```diff
@@ -9,17 +9,25 @@
...
 /**
  * Base class for the group folder expiration commands.
  */
 class ExpireGroupBase extends Base {
+	/**
+	 * @throws LogicException
+	 */
 	public function __construct() {
 		parent::__construct();
 	}
```
come-nc, line 26:
```suggestion
```
dead code

come-nc (commented):
I kind of hate that `Exception` is sometimes `\Exception` and sometimes `\OCP\DB\Exception`. I tend to never `use Exception` because of that to make it clear when we throw the root one.

Also `@throws` blocks should contain an explanation when the exceptions is thrown, except for self-explanatory ones like `NoUserException` and `InvalidArgumentException`.

And for all methods implementing an OCP interface they should not throw unless it’s permitted by the interface. But that does mean for some of them we should modify the interface instead.
For controllers I do not know the policy, are they allowed to throw, and which kind of exceptions/throwable is handled by Nextcloud core?

PR #3323, opened 2024-10-07
fix(BackgroundJob): Set expire job intervals to 30 minutes

provokateurin:
To match the server intervals:
https://github.com/nextcloud/server/blob/363ba6b0f5b99583e789695f19dcd96db00508ef/apps/files_trashbin/lib/BackgroundJob/ExpireTrash.php#L31
https://github.com/nextcloud/server/blob/363ba6b0f5b99583e789695f19dcd96db00508ef/apps/files_versions/lib/BackgroundJob/ExpireVersions.php#L27

### DIFF
--- a/lib/BackgroundJob/ExpireGroupTrash.php
+++ b/lib/BackgroundJob/ExpireGroupTrash.php
@@ -22,8 +22,8 @@ class ExpireGroupTrash extends TimedJob {
 		ITimeFactory $timeFactory,
 	) {
 		parent::__construct($timeFactory);
-		// Run once per hour
-		$this->setInterval(60 * 60);
+		// Run every 30 minutes
+		$this->setInterval(60 * 30);
 	}
 
 	protected function run(mixed $argument): void {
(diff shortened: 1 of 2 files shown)


come-nc:
Why not align server instead? Or why align at all?

30minutes seems really often for most setups.
If this is because of users setting expiration to less that 1hour, maybe set a shorter interval only in this edge case?

provokateurin:
> Why not align server instead?

Could also do that, to me aligning Groupfolders feels more natural because server is kind of a source of truth (especially given the fact that there is no explanation for any of the intervals anywhere).

> Or why align at all?

Because it is easier for admins if all trash/versions expiry jobs have the same interval. This came up in a recent support case as you can see in the files team chat messages from this morning.

> 30minutes seems really often for most setups.

Agreed, but I don't know the reasoning for any of these intervals as they were never changed after the initial implementation.

> If this is because of users setting expiration to less that 1hour, maybe set a shorter interval only in this edge case?

But this doesn't make sense either, then it should be the lowest possible interval to ensure the files are deleted ASAP.


I'm not sure if there is anyone who can actually shed some light onto the situation.

PR #3324, opened 2024-10-07
feat: Add OpenAPI

provokateurin:
Closes https://github.com/nextcloud/groupfolders/issues/3208
~This is a "breaking" change to the API, but since it was not usable before due to CSRF restrictions it doesn't matter.
While working on this I also had to cleanup a lot of the internal typings which resulted in this relatively large change for adding OpenAPI.
If this is too big please let me know and I will try to split it up.~


README.md
```diff
@@ -126,35 +126,7 @@ To disable the advanced permissions feature for a group folder, use `occ groupfo
...
-- `POST apps/groupfolders/folders/$folderId/quota`: Set the quota for a folder
-    - `quota`: The new quota for the folder in bytes, user `-3` for unlimited
-- `POST apps/groupfolders/folders/$folderId/mountpoint`: Change the name of a folder
-    - `mountpoint`: The new name for the folder
-
-For all `POST` calls the required parameters are listed.
-
-Non-admins can access the `GET` requests to retrieve info about group folders they have access to.
-Admins can add `applicable=1` as a parameter to the group folder list request to get the same filtered results of only folders they have direct access to.
+See the [OpenAPI specification](openapi.json) to learn about all available API endpoints.
```
come-nc, line 129: Hum any chance at linking to a readable form of the doc?

lib/Command/ListCommand.php
```diff
@@ -87,8 +87,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 
 		if ($outputType === self::OUTPUT_FORMAT_JSON || $outputType === self::OUTPUT_FORMAT_JSON_PRETTY) {
 			foreach ($folders as &$folder) {
-				$folder['group_details'] = $folder['groups'];
```
come-nc, line 90: Why do you remove group_details? This is a breaking change for any script relying on that.

lib/Command/ListCommand.php
```diff
@@ -98,13 +101,17 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+				if (is_array($folder['groups'])) {
+					$groupStrings = array_map(function (string $groupId, array $entry) use ($groupNames): string {
+						[$permissions, $displayName] = [$entry['permissions'], $entry['displayName']];
+						$groupName = array_key_exists($groupId, $groupNames) && ($groupNames[$groupId] !== $groupId) ? $groupNames[$groupId] . ' (' . $groupId . ')' : $displayName;
+
+						return $groupName . ': ' . $this->permissionsToString($permissions);
+					}, array_keys($folder['groups']), array_values($folder['groups']));
+					$folder['groups'] = implode("\n", $groupStrings);
+				} else {
+					$folder['groups'] = [];
```
come-nc, line 113:
```suggestion
					$folder['groups'] = '';
```

lib/Folder/Folder.php
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\GroupFolders\Folder;
+
+use OC\Files\Cache\CacheEntry;
+
+class Folder {
```
come-nc, line 14: Avoid using the same name as core classes, it’s confusing when the class is imported with a `use` statement.

lib/Folder/FolderManager.php
```diff
@@ -59,10 +69,14 @@ public function getAllFolders(): array {
 		$folderMap = [];
 		foreach ($rows as $row) {
 			$id = (int)$row['folder_id'];
-			$folderMap[$id] = [
+			$applicables = $applicableMap[$id] ?? [];
+			if (empty($applicables)) {
+				$applicables = new stdClass();
```
come-nc, line 74: This is a controller specific thing that ends up in the manager, is there no way to avoid that?

lib/Folder/FolderMapping.php
```diff
@@ -0,0 +1,19 @@
...
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\GroupFolders\Folder;
+
+class FolderMapping {
```
come-nc, line 12: These classes should actually be entities with mappers, no?

lib/Command/ListCommand.php
provokateurin: Yes, as I said at the top.
This was discussed in last weeks team call with @sorbaugh.
It is fine to do this once for the benefit of a much cleaner API.

README.md
provokateurin: Not sure how that would work since we do not have an existing documentation page for Groupfolders.

artonge: Can you generate a `.md` files from the openapi doc? Then we could link to that?

lib/Folder/FolderMapping.php
provokateurin: I guess so. I only made this change from array to class last minute, so I didn't give it much thought.

lib/Folder/Folder.php
provokateurin: But `Groupfolder` is also not a good name. Would `InternalFolder` be better as it is only used as an internal type between different methods?

provokateurin:
I will try to extract some of the changes into separate PRs to get them merged first and then come back to this PR.

lib/Folder/Folder.php
come-nc: As said below this should be in Db and be an entity with a mapper ideally.
I do think Groupfolder is closer to what it is than Folder.
Could be FolderDefinition or FolderSettings if that makes sense?

README.md
come-nc: I was thinking of a magic link like https://petstore.swagger.io/?url=https://raw.githubusercontent.com/nextcloud/groupfolders/ac3b696dbb820a5ae34039be1300c1309e58d4e1/openapi.json
tables application does that.
Using this instance is not ideal as it’s supposed to be a demo and it may even be contrary to their ToS, I did not check. But maybe there is a similar tool we can use, or we can host our own?

lib/Folder/FolderMapping.php
provokateurin: Moved to https://github.com/nextcloud/groupfolders/issues/3389, I will revert these changes in this PR

provokateurin:
I removed everything not strictly necessary to implement this. Some improvements can just be done later, like https://github.com/nextcloud/openapi-extractor/issues/172.

lib/Folder/FolderManager.php
```diff
@@ -489,19 +514,19 @@ public function getFoldersForGroup(string $groupId, int $rootStorageId = 0): arr
 
 		$result = $query->executeQuery()->fetchAll();
 
-		return array_values(array_map(fn (array $folder): array => [
+		return array_values(array_values(array_map(fn (array $folder): array => [
```
come-nc, line 517: A bit too much? :open_mouth:

come-nc (commented):
Why removing all the id indexing, is that so bad?
Also, why the snake_case on string ids, do we have a convention about that in OCS?

provokateurin:
> Why removing all the id indexing, is that so bad?

Not really, but I just removed it because it is never used.
I prefer a `list` strongly over `array` if possible, due to the ambiguities that can happen.
I can also revert it, if you really don't like it.

>  Also, why the snake_case on string ids, do we have a convention about that in OCS?

No, I just wanted to make them all consistent.
I personal prefer snake_case for JSON, but we can also go with camelCase.

susnux:
> I personal prefer snake_case for JSON, but we can also go with camelCase.

I think we use camelCase for both PHP and JS variables and for JS working with camelCase JSON is more convenient. So if we somewhen might consider a codestyle also JSON camelCase would be nice

provokateurin:
> I think we use camelCase for both PHP and JS variables and for JS working with camelCase JSON is more convenient. So if we somewhen might consider a codestyle also JSON camelCase would be nice

Ok, fair enough :)

provokateurin:
> Why removing all the id indexing, is that so bad?

Also the code on the frontend side was very hacky in this regard, it is a lot better now.

provokateurin:
Depends on https://github.com/nextcloud/groupfolders/pull/3407 and https://github.com/nextcloud/openapi-extractor/pull/181.
No breaking changes anymore, but also very ugly typing :see_no_evil:

provokateurin:
All depending PRs are merged, so this is finally ready.

sorbaugh (approved):

provokateurin:
Still waiting for a review from @come-nc before merging to ensure I addressed all previous problems.

provokateurin:
Thanks a lot everyone! :tada:

PR #3325, opened 2024-10-08
fix(FolderManager): Use placeholder value for default quota instead of the current value on creation

provokateurin:
Follow-up to https://github.com/nextcloud/groupfolders/pull/3160

### DIFF
--- a/tests/Folder/FolderManagerTest.php
+++ b/tests/Folder/FolderManagerTest.php
@@ -39,10 +39,6 @@ class FolderManagerTest extends TestCase {
 		$this->logger = $this->createMock(LoggerInterface::class);
 		$this->eventDispatcher = $this->createMock(IEventDispatcher::class);
 		$this->config = $this->createMock(IConfig::class);
-		$this->config->expects($this->any())
-			->method('getSystemValueInt')
-			->with('groupfolders.quota.default', FileInfo::SPACE_UNLIMITED)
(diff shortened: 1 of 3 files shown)


tests/Folder/FolderManagerTest.php
```diff
@@ -399,4 +447,20 @@ public function testGetFolderPermissionsForUserMerge(): void {
...
+		$this->config->expects($this->any())
+			->method('getSystemValueInt')
+			->with('groupfolders.quota.default', FileInfo::SPACE_UNLIMITED)
+			->willReturnCallback(function () use (&$exponent) {
+				return 1024 ** ($exponent++);
+			});
+
+		$this->assertEquals(1024 ** 3, $this->manager->getFolder($folderId1, $rootStorageId)['quota']);
+		$this->assertEquals(1024 ** 4, $this->manager->getFolder($folderId1, $rootStorageId)['quota']);
+	}
```
provokateurin, line 465: For some reason adding this test makes the trash tests fail. Keeping everything the same and just removing this test makes them work again.
I tried to debug this for a while, but couldn't make any sense of it.
The difference for the trash tests is that the permissions get calculated differently because a rule is not added which causes the tests to fail as they expect the normal user to no longer see the file without read permission.

come-nc (approved):

provokateurin:
(CI failure was only due to https://github.com/nextcloud/server/pull/48610)

PR #3326, opened 2024-10-08
fix: update parent etag when changing acl rules

icewind1991:
This ensures that sync clients can pickup on the fact that a user might have gained/lost access to a file.

It also updates the etag for users that don't have any changes, but it's better to be on the safe side here and figuring out which users are affected would add significant complexity.

### DIFF
--- a/lib/DAV/ACLPlugin.php
+++ b/lib/DAV/ACLPlugin.php
@@ -236,6 +236,9 @@ class ACLPlugin extends ServerPlugin {
 				$this->ruleManager->saveRule($rule);
 			}
 
+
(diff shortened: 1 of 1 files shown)


come-nc (approved):

come-nc:
Maybe add a comment before the line to explain

provokateurin (approved):

provokateurin:
I think there was an issue about this problem but I can't find it anymore.

provokateurin:
Ah found it: https://github.com/nextcloud/desktop/issues/7150

mgallien:
tested with desktop client
works fine and allows it to detect changes in permissions when files themselves do not change
thanks @icewind1991

PR #3328, opened 2024-10-09
fix: Override check for applying quota in wrapper

juliushaertl:
Otherwise the parent Quota class takes over and blocks applying the quota

Requires https://github.com/nextcloud/server/pull/48623

But should not break when updating groupfolders without the server patch: https://3v4l.org/eekkT


provokateurin (approved):

provokateurin:
Would be nice to run the CI after the server PR is merged to ensure none of the tests break.

Issue #3339, opened 2024-10-10
Restoring of deleted shared file inside folder inside groupfolder not possible


x7airworker:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Create a groupfolder "A" with ACL and give readonly permissions to user "A"
2. Create a folder "B" inside groupfolder "A" and give full access to user "A"
3. Share the folder with full permissions to another user "B"
4. Create a file inside "C" and share it with another user "B"
5. Login to the other user "B" and delete file "C"
6. Try to restore the file with any of the users "A" or "B".

### Expected behaviour
The file should be able to be restorable, with at least user "A".

### Actual behaviour
The request results in http status 500 with a NotPermittedException.
I've noticed that the `original_location` column in table `oc_group_folders_trash` doesn't contain the folder B, but just the name of file C with the folder_id of groupfolder A. I think this results in a bad ACL check.

### Server configuration

**Operating system:** Alpine; Docker

**Web server:** Nginx

**Database:** MySQL (Galera)

**PHP version:** 8.2.7

**Nextcloud version:** 28.0.9

**Group folders version:** 16.0.8

**Updated from an older Nextcloud/ownCloud or fresh install:** fresh install

**Where did you install Nextcloud from:** official source; self-built docker image

**Are you using external storage, if yes which one:** s3

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** no

### Client configuration
**Browser:** Chrome

**Operating system:** MacOS 15

#### Web server error log
<details>
<summary>Web server error log</summary>

```
127.0.0.1 -  10/Oct/2024:08:22:58 +0000 "MOVE /remote.php" 500
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
[...]

PR #3358, opened 2024-10-14
fix(Trash): Fix original location for deleting shared item

provokateurin:
Fixes https://github.com/nextcloud/groupfolders/issues/3339

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -24,6 +24,7 @@ use OCP\Files\IRootFolder;
 use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
+use OCP\Files\Storage\ISharedStorage;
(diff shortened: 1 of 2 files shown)


provokateurin:
I'm not sure if there is a more common scenario to trigger this, so that the fix needs to be extended to other storages as well.

come-nc (approved):

x7airworker:
@provokateurin Thanks for the fix! Would it be possible to also get a backport to stable28? :)

provokateurin:
No, unfortunately 28 is no longer supported at all. You should update to at least 29 to still receive security and bug fixes.

For some reasons the tests don't pass on 29 and 30, so I'm also not going to fix it on those versions and it will only be in the new 31 version that will be released in less than a month.
If you want you can try to find a fix for 29 and 30 (I expect them to have the same problem) and I'm happy to review those PRs if you manage to do it.

PR #3362, opened 2024-10-15
fix: include mountpoint in original location for trash items

icewind1991:
Currently the "original location" for deleted groupfolder items don't show the path of the groupfolder itself. Making it harder for users to know what the deleted file is.

This changes it to store the "full" original location and separately keep the "interal" original location for the restore operation.


icewind1991:
This (should) only changes how the path is displayed. So I think it's orthogonal from that issue/test, which I understand to be caused by what path is stored.

provokateurin:
You're right, I confused this with another issue.

provokateurin (approved):
Makes sense :+1:

PR #3369, opened 2024-10-18
refactor: Use constructor property promotion where possible

provokateurin:


lib/ACL/ACLStorageWrapper.php
```diff
@@ -176,10 +175,8 @@ public function writeStream(string $path, $stream, ?int $size = null): int {
 
 	/**
 	 * @inheritDoc
-	 * @param string $path
-	 * @param ?IStorage $storage
 	 */
-	public function getCache($path = '', $storage = null): ICache {
+	public function getCache(string $path = '', ?\OCP\Files\Storage\IStorage $storage = null): ICache {
```
come-nc, line 179: We should apply the auto-import as in server first, to avoid it removing the `use`.

provokateurin:
Cypress failure is unrelated, same errors are on master as well :/

come-nc (approved):

PR #3390, opened 2024-10-29
perf: Add index for group_folders table

provokateurin:
We do a lot of SELECTs on this table in the FolderManager, but there is actually no index on it at all. All SELECTs include all columns, so that's why I'm not adding one for folder_id only.
Usually not noticable when there aren't many groupfolders, but if you got a few then it can take a bit too long.

### DIFF
--- a/lib/Migration/Version19000Date20241029123147.php
+++ b/lib/Migration/Version19000Date20241029123147.php
@@ -16,10 +16,7 @@ use OCP\Migration\SimpleMigrationStep;
 
 class Version19000Date20241029123147 extends SimpleMigrationStep {
 	/**
-	 * @param IOutput $output
 	 * @param Closure(): ISchemaWrapper $schemaClosure
-	 * @param array $options
-	 * @return null|ISchemaWrapper
 	 */
 	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
 		/** @var ISchemaWrapper $schema */
@@ -29,9 +26,10 @@ class Version19000Date20241029123147 extends SimpleMigrationStep {
 			$table = $schema->getTable('group_folders');
 			if (!$table->hasIndex('gf_folders_folder_id')) {
 				$table->addUniqueIndex(['folder_id'], 'gf_folders_folder_id');
+				return $schema;
 			}
 		}
 
-		return $schema;
+		return null;
 	}
 }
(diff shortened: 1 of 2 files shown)


lib/Migration/Version19000Date20241029123147.php
```diff
@@ -0,0 +1,35 @@
...
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('group_folders');
+		if (!$table->hasIndex('group_folders_folders')) {
+			$table->addIndex(['folder_id', 'mount_point', 'quota', 'acl'], 'group_folders_folders');
```
artonge, line 30: Including those 4 columns means that we need to have them in the request conditions. Are all requests following that format?

```sql
SELECT ...
WHERE
  folder_id = ... AND
  mount_point = ... AND
  quota = ... AND
  acl = ...
```

provokateurin: Eh you're right, I confused it with the selected columns...

artonge (approved):

artonge (commented):
I doubt it would improve anything given that I don't see how the number of groupfolders could be higher than 1000. And with that amount, an index would hardly make a difference, no?

But I guess it does not hurt

provokateurin:
Not substantial performance improvements measured, we can still add it later if needed.

icewind1991:
There are instances with 10k+ groupfolders so having an index probably makes sense.

provokateurin:
Ok, then we can add it. I would have expected that there already is an index in case anyone already suffered performance problems from the lack of an index on the table.

lib/Migration/Version19000Date20241029123147.php
```diff
@@ -0,0 +1,37 @@
...
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('group_folders')) {
+			$table = $schema->getTable('group_folders');
+			if (!$table->hasIndex('gf_folders_folder_id')) {
+				$table->addUniqueIndex(['folder_id'], 'gf_folders_folder_id');
+			}
+		}
+
+		return $schema;
```
come-nc, line 35: You should return $schema in the inner `if` and return `null` in other cases.

come-nc (changes requested):

come-nc (approved):

PR #3393, opened 2024-10-31
Fix labeled version autoExpire

XueSheng-GIT:
Fix https://github.com/nextcloud/groupfolders/issues/3095

Follow up to https://github.com/nextcloud/groupfolders/pull/3213 which did not consider labeled versions of `getAutoExpireList`.

### DIFF
--- a/lib/Versions/ExpireManager.php
+++ b/lib/Versions/ExpireManager.php
@@ -66,8 +66,11 @@ class ExpireManager {
 			while ($newInterval) {
 				if ($nextInterval === -1 || $prevTimestamp > $nextInterval) {
 					if ($version->getTimestamp() > $nextVersion) {
-						//distance between two version too small, mark to delete
-						$toDelete[] = $version;
+						// Do not expire versions with a label.
+						if (!($version instanceof IMetadataVersion) || $version->getMetadataValue('label') === null || $version->getMetadataValue('label') === '') {
+							//distance between two version too small, mark to delete
+							$toDelete[] = $version;
+						}
 					} else {
 						$nextVersion = $version->getTimestamp() - $step;
 						$prevTimestamp = $version->getTimestamp();


lib/Versions/ExpireManager.php
```diff
@@ -66,8 +66,11 @@ protected function getAutoExpireList(int $time, array $versions): array {
 			while ($newInterval) {
 				if ($nextInterval === -1 || $prevTimestamp > $nextInterval) {
 					if ($version->getTimestamp() > $nextVersion) {
-						//distance between two version too small, mark to delete
-						$toDelete[] = $version;
+						// Do not expire versions with a label.
+						if ($version->getMetadataValue('label') === null || $version->getMetadataValue('label') === '') {
```
artonge, line 70: Should never occur, but makes it obvious and makes psalm happy

```suggestion
						if (!($version instanceof IMetadataVersion) || $version->getMetadataValue('label') === null || $version->getMetadataValue('label') === '') {
```

artonge: Not sure how it fits with the 'auto expire' logic. This probably retain more version than wanted, but prevents data loss. So fine with me.

XueSheng-GIT: Depending on the specific case, more versions will be retained. But that's the intention if a version is labeled, right? But expectations could be different.
Actually, while doing initial tests with this additional check for `IMetadataVersion`, versions were not retained as expected. Thus, I will do some more tests.

XueSheng-GIT:
@artonge do I need to squash the commits or is it fine like that?

artonge:
> @artonge do I need to squash the commits or is it fine like that?

Yes please :)

artonge (approved):

XueSheng-GIT:
@artonge anything else I need to do to get this fix merged?

PR #3406, opened 2024-11-11
test: Fix cypress tests

provokateurin:
I still see some failures locally, but I think they might just be local problems that are not going to happen in CI :woman_shrugging:

I updated the filesUtils.ts from server and also migrated some of the tests to use these helpers where possible to ease maintaining them in the future.


provokateurin:
Ok, the preview icon errors seem to be real problems, not only locally.

sorbaugh (approved):

PR #3433, opened 2024-11-18
ignore inShare from cli

ArtificialOwl:
because `::currentUID()` is null when executed from occ...

This fix an issue that block access to documents if ACL denies Sharing [1], even if read/write/delete permissions are fine, when calling `IUserFolder()->getbyId();`

https://github.com/nextcloud/groupfolders/blob/master/lib/ACL/ACLCacheWrapper.php#L35


susnux (approved):

come-nc:
This inShare stuff is weird

Issue #3409, opened 2024-11-11
Group filter input field characters are not visible (in groupfolder admin settings when selecting group for the groupfolder)


andrey18106:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Create test groups of users
2. Create test groupfolder
3. Open "Add group or team"
4. Type the group name to filter
5. See no characters visible, but filtering works

![Image](https://github.com/user-attachments/assets/0d87c96b-7e74-4658-bf8d-1244cb182967)

### Expected behaviour
The input group name is visible in the input

### Actual behaviour
The input group name is not visible.

### Server configuration

**Operating system:**

**Web server:**

**Database:**

**PHP version:**

**Nextcloud version:** (see Nextcloud admin page)

**Group folders version:**

**Updated from an older Nextcloud/ownCloud or fresh install:**

**Where did you install Nextcloud from:**

**Are you using external storage, if yes which one:** local/s3/smb/sftp/...

**Are you using encryption:** yes/no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/Saml/...

### Client configuration
**Browser:**

**Operating system:**

#### Web server error log
<details>
<summary>Web server error log</summary>

```
Insert your webserver log here
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
Insert your Nextcloud log here
```
</details>

#### Browser log
<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```
</details>

PR #3460, opened 2024-12-04
fix(settings): Fix select input color

provokateurin:
Fixes https://github.com/nextcloud/groupfolders/issues/3409


provokateurin:
The library has it's own styling which interferes with our styling, so we have to manually set the text color in order for it to be legible on the background.

Issue #3524, opened 2025-01-07
String rename "Groupfolders" to "Team folders"


sorbaugh:
To bring Groupfolders more in line with Teams (previously Circles), here too, should Groupfolders be renamed to "Team folders".

Note that this should affect strings only (no repository or endpoint changes).

solracsf:
Please check this RFE too: https://github.com/nextcloud/groupfolders/issues/3494

PR #3536, opened 2025-01-13
chore: Rename app to Team folders

provokateurin:
Closes https://github.com/nextcloud/groupfolders/issues/3524

I hope I found all the user-facing strings.
I also made the names for consistent, so all variations of "groupfolders", "group folders", "Group folders" and "Groupfolders" are just "Team folders" now.

### DIFF
--- a/openapi.json
+++ b/openapi.json
@@ -3,7 +3,7 @@
     "info": {
         "title": "groupfolders",
         "version": "0.0.1",
-        "description": "Admin configured folders shared with everyone in a group",
+        "description": "Admin configured folders shared with everyone in a team",
         "license": {
             "name": "agpl"
         }
--- a/.github/ISSUE_TEMPLATE/Bug_report.md
+++ b/.github/ISSUE_TEMPLATE/Bug_report.md
@@ -7,7 +7,7 @@ labels: bug, 0. Needs triage
 <!--
 Thanks for reporting issues back to Nextcloud!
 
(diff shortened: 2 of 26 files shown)


openapi.json
```diff
@@ -3,7 +3,7 @@
     "info": {
         "title": "groupfolders",
         "version": "0.0.1",
-        "description": "Admin configured folders shared with everyone in a group",
+        "description": "Admin configured folders shared with everyone in a Team",
```
come-nc, line 6: These changes are ambiguous, because circles are called Teams now, so this make it sound like it’s specific to teams/circles, no?

provokateurin: We have the same situation in other places as well.
I think calling it `group/Team` or `group or Team` makes it harder to understand.
So I would try to use `Team` everywhere, except where we really mean `group` on a technical level.

provokateurin: I asked @marcoambrosini and he seems to agree, so let's keep it like it is now.

icewind1991 (approved):

jancborchardt (dismissed):
Looks good, just 1 detail: Unless it’s at the beginning of a sentence, "team" should be lowercase. It’s a basic English word and looks out of place otherwise. :)

provokateurin:
Ok, I thought we were taking it as a proper noun like in "Team folder" :+1:

sorbaugh (approved):

solracsf:
Just like the Circles (Teams) app, a note about the rename in the Readme.md can't hurt:

> The app was formerly named circles which is still used as internal identifier a lot.

Ornanovitch:
This is huge, what is the project behind this? Is there any issue tracking some sort of unification of groups and teams? If it's only a rebranding without any other meaning, there is a massive risk of confusion in the NC ecosystem...

provokateurin:
The reason is that Circles was also renamed to Teams and the names should be aligned, but @sorbaugh or @jancborchardt should be able to tell you more about this.

marcoambrosini:
@Ornanovitch this is the general idea behind it https://github.com/nextcloud/server/issues/49488

PR #3551, opened 2025-01-23
include circles to acl

ArtificialOwl:
This adds Teams to the ACL management:

- search will now returns Teams and Teams' members so that Circles can be added to the list of ACL Manager in the admin settings of the app,
- ACL rights about Teams and Teams' members can be set in the right panel when navigating a groupfolders' files,
- ACL set regarding a Teams and Teams' members are applied to files and folders.

Missing front-end:
- [x] In the admin part, while the Teams' member are displayed, the Teams are missing in the dropdown list when adding ACL Manager. This is due to the fact that the data stored behind the array key 'circles' are not managed by the javascript.
- [x] In the admin part, when displaying a Team, it is shown as 'Group' instead of 'Team'
- [x] In the Files' right panel, while the Teams' member are displayed, the Teams are not in the dropdown list when adding ACL Permissions.

### DIFF
--- a/lib/ACL/UserMapping/UserMappingManager.php
+++ b/lib/ACL/UserMapping/UserMappingManager.php
@@ -8,33 +8,109 @@ declare(strict_types=1);
 
 namespace OCA\GroupFolders\ACL\UserMapping;
 
+use OCA\Circles\CirclesManager;
+use OCA\Circles\Exceptions\CircleNotFoundException;
+use OCA\Circles\Model\Circle;
+use OCA\Circles\Model\Probes\CircleProbe;
+use OCP\AutoloadNotAllowedException;
 use OCP\IGroup;
 use OCP\IGroupManager;
 use OCP\IUser;
 use OCP\IUserManager;
+use OCP\Server;
+use Psr\Container\ContainerExceptionInterface;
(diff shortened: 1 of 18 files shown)


lib/ACL/UserMapping/UserMappingManager.php
```diff
@@ -8,33 +8,78 @@
...
-		} else {
+		switch ($type) {
+			case 'group':
+				$displayName = $this->groupManager->get($id)?->getDisplayName();
+				break;
+			case 'user':
+				$displayName = $this->userManager->get($id)?->getDisplayName();
+				break;
+			case 'circle':
+				$displayName = 'qwerty';
```
icewind1991, line 48: placeholder :see_no_evil:

icewind1991 (approved):
:+1: for @ArtificialOwl's part

skjnldsv (approved):
Code looks sane 💅

AndyScherzinger (approved):

icewind1991:
I'll do further backports from the 30 backport to reduce the conflict resolution work

icewind1991:
29: qemu-nbd --disconnect /dev/nbd0

PR #3566, opened 2025-01-27
chore(deps): Update psalm to v6

provokateurin:
The two new entries to the baseline are false positives IMO, because psalm has the wrong signature without the callable that can be seen in the docs: https://www.php.net/manual/en/function.array-udiff.php


tests/psalm-baseline.xml
artonge, line 1: Can't we have the exception in the code instead of in the baseline?

provokateurin: I think the baseline is a better fit, in case this false positive gets fixed in a later psalm version. Otherwise we have an ignore in the code which would become useless. It's not like we intentionally want to ignore it because it is correct, we ignore it because psalm is broken.

PR #3570, opened 2025-01-27
refactor: Apply rector refactorings

provokateurin:


lib/Command/ACL.php
```diff
@@ -117,7 +117,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$mappingType = $input->getOption('user') ? 'user' : 'group';
 			$mappingId = $input->getOption('user') ?: $input->getOption('group');
 			$path = $input->getArgument('path');
-			$path = trim($path, '/');
+			$path = trim((string)$path, '/');
```
artonge, line 120: Maybe instead of casting we should check the content?
What the current type here? `string|null`? Is `(string)null === ''`? If so, then all good.

lib/Command/ACL.php
```diff
@@ -152,7 +152,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 					return -3;
 				}
 
-				$name = substr($permission, 1);
+				$name = substr((string)$permission, 1);
```
artonge, line 155: Same as above

lib/Command/ACL.php
```diff
@@ -221,7 +221,7 @@ private function parsePermissions(array $permissions): array {
 		$result = 0;
 
 		foreach ($permissions as $permission) {
-			$permissionValue = Rule::PERMISSIONS_MAP[substr($permission, 1)];
+			$permissionValue = Rule::PERMISSIONS_MAP[substr((string)$permission, 1)];
```
artonge, line 224: Sam as above

lib/Command/Quota.php
```diff
@@ -29,7 +29,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			return -1;
 		}
 
-		$quotaString = strtolower($input->getArgument('quota'));
+		$quotaString = strtolower((string)$input->getArgument('quota'));
```
artonge, line 32: Same

artonge (approved):
Looks good, besides potentially ignorable comments about the castings.

lib/Mount/CacheRootPermissionsMask.php
```diff
@@ -15,14 +15,14 @@
...
 	public function __construct(
 		ICache $cache,
-		private int $mask,
+		private readonly int $mask,
 	) {
 		parent::__construct($cache);
 	}
 
 	protected function formatCacheEntry($entry): ICacheEntry|false {
 		$path = $entry['path'];
```
come-nc, line 24:
```suggestion
		$path = (string)$entry['path'];
```
It’s better to cast early rather than when the variable is used.
It’s the same for most of the other cast changes actually.

tests/Folder/FolderManagerTest.php
```diff
@@ -61,7 +61,7 @@ private function clean(): void {
 	private function assertHasFolders(array $folders): void {
 		$existingFolders = array_values($this->manager->getAllFolders());
 		usort($existingFolders, fn (array $a, array $b): int => strcmp($a['mount_point'], $b['mount_point']));
-		usort($folders, fn (array $a, array $b): int => strcmp($a['mount_point'], $b['mount_point']));
+		usort($folders, fn (array $a, array $b): int => strcmp((string)$a['mount_point'], (string)$b['mount_point']));
```
come-nc, line 64: Why this one and not the line just before :thinking:

tests/ACL/ACLManagerTest.php
```diff
@@ -51,11 +51,7 @@ protected function setUp(): void {
 			});
 
 		$this->ruleManager->method('getRulesForPrefix')
-			->willReturnCallback(function (IUser $user, int $storageId, string $prefix) {
-				return array_filter($this->rules, function (string $path) use ($prefix) {
-					return $prefix === $path || str_starts_with($path, $prefix . '/');
-				}, ARRAY_FILTER_USE_KEY);
-			});
+			->willReturnCallback(fn (IUser $user, int $storageId, string $prefix): array => array_filter($this->rules, fn (string $path): bool => $prefix === $path || str_starts_with($path, $prefix . '/'), ARRAY_FILTER_USE_KEY));
```
artonge, line 54: Seems less readable than the previous version.

lib/Command/ACL.php
```diff
@@ -117,8 +117,16 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$mappingType = $input->getOption('user') ? 'user' : 'group';
 			$mappingId = $input->getOption('user') ?: $input->getOption('group');
 			$path = $input->getArgument('path');
+			if (!is_string($path)) {
+				$output->writeln('<error><path> argument has to be a string</error>');
+				return -3;
```
come-nc, line 122: `-3`?
Why?
Maybe `self::INVALID`? (would be `2`)

provokateurin: I just copied this behavior from other checks above :woman_shrugging:

come-nc (approved):

PR #3735, opened 2025-04-08
ACL on nested teams

### LINKED ISSUE #3722: Nested Teams do not work when adding an advanced permission rule
Hello!
It is very nice that it is now possible to use Teams in advanced permissions. This could allow delegating access control rights to nested folders to non-administrator users.
But there is still a problem. In the normal permissions of the group folder "Group or Team", you need to explicitly specify all Teams that can have access to the group folder and nested folders. Only an administrator can do this.
A solution could be to create one root Team, and non-administrator users could add the Teams they created there to restrict access.
I tried to do this, but when assigning advanced permissions, the nested Teams are simply not visible.

Reproducing the problem:

1. Create the Team "Root_Team"
2. Create the Team "Team_1"
3. Create the Team "Team_2"
4. Include Team_1and Team_2 in "Root_Team"
5. Create the Group Folder "Folder_1"
6. Assign the Team "Root_Team" to the folder in "Group or Team"
7. Go to "Files" and try to assign an advanced permission rule to the Folder "Folder_1".
8. The Teams Team_1 and Team_2 are not in the selection.

ArtificialOwl:
- When creating a new Team Folders as admin, all teams are visible (and can be selected/added to the team folder) disregard being a member of those teams
- Nested teams can now be selected to be set as Advanced Permissions Manager in the Admin Settings,
- ACL for a nested team can be set in the right panel/Sharing tab

should fix #3722

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -432,10 +432,26 @@ class FolderManager {
 		$circles = $this->getAllApplicable()[$id] ?? [];
 		$circles = array_map(fn (string $singleId): ?Circle => $this->getCircle($singleId), array_keys($circles));
 
+		// get nested teams
+		$nested = [];
+		foreach ($circles as $circle) {
+			try {
+				$inherited = $circle?->getInheritedMembers(true) ?? [];
+			} catch (\Exception $e) {
+				$this->logger->notice('could not get nested teams', ['exception' => $e]);
+				continue;
+			}
+			foreach ($inherited as $entry) {
+				if ($entry->getUserType() === Member::TYPE_CIRCLE) {
+					$nested[] = $entry->getBasedOn();
+				}
+			}
+		}
+
 		return array_map(fn (Circle $circle): array => [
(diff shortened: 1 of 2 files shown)


lib/Folder/FolderManager.php
```diff
@@ -431,10 +431,26 @@ private function getCircles(int $id): array {
...
+			} catch (\Exception $e) {
+				$this->logger->notice('could not get nested teams', ['exception' => $e]);
+				continue;
+			}
+			foreach ($inherited as $entry) {
+				if ($entry->getUserType() === Member::TYPE_CIRCLE) {
+					$nested[] = $entry->getBasedOn();
+				}
+			}
+		}
```
provokateurin, line 448: Does this properly account for triple nested teams as well?

ArtificialOwl: Yes, inheritedmembers is using memberships table and will returns all nested memberships

provokateurin (approved):

PR #3751, opened 2025-04-24
fix(dav): only list top-level groupfolders

skjnldsv:
You can have multiple folders with the same name
You can also have many of those same names nested into other sub-groupfolders.
Example:

```
├── TeamFolder1
│   ├── TeamFolder1
│   ├── TeamFolder2
│   ├── TeamFolder3
│   ├── TeamFolder4
│   └── TeamFolder5
└── TeamFolder2
    ├── TeamFolder1
    ├── TeamFolder2
    ├── TeamFolder3
    ├── TeamFolder4
    └── TeamFolder5
```

## Resolution:
1. We now only show top-level groupfolders, as clicking one will show the children anyway
2. We now group them by source name, and use the groupfolder ID instead of the mountpoint to generate their source. Making them truly unique

### DIFF
--- a/lib/DAV/GroupFoldersHome.php
+++ b/lib/DAV/GroupFoldersHome.php
@@ -103,6 +103,12 @@ class GroupFoldersHome implements ICollection {
 		}
 
 		$folders = $this->folderManager->getFoldersForUser($this->user, $storageId);
+
+		// Filter out top-level folders only
+		$folders = array_filter($folders, function (array $folder) {
+			return !str_contains($folder['mount_point'], '/');
+		});
+
 		return array_map($this->getDirectoryForFolder(...), $folders);
 	}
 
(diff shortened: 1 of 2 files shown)


skjnldsv:
@provokateurin yes and no, there are multiple endpoints used here.
This one is the dedicated groupfolder listing endpoint, so a fix on the front is actually needed anyway.

Also, while I agree we need a backend fix, I cannot wait for a deeper refactor of our code.
Mitigating the issue is the priority

lib/DAV/GroupFoldersHome.php
```diff
@@ -103,6 +103,12 @@ public function getChildren(): array {
 		}
 
 		$folders = $this->folderManager->getFoldersForUser($this->user, $storageId);
+
+		// Filter out top-level folders only
```
artonge, line 107:
```suggestion
		// Filter out non top-level folders
```

artonge (approved):

Issue #681, opened 2019-11-22
duplicate group folder names in the nextcloud group folder app


mesiah1970-haydar:
Hello and thanks,

here is my process description of group folders.

I don’t want to have the users create their own folders. So I keep their contingent at 0 Bytes.
This works perfect. So they are only able to see the folders, which are shared over the group folders app. So no personal folders.

I create a user group, let’s say customername.projectname
I add a project user to the newly created group
I go to group folders settings in the admin panel.
I create a new folder like customername.projectname and add the newly created group.
Now the issue is, that I’m able to create the same foldername again an again again …
So this is really misleading. The System should tell me, that the folder with the same name already exists, not more. Like in Unix or Windows. This is a basic requirement, which does not meet in group folders.

Frankly I could not understand your technical solution exactly, could please explain it in detail.

Many thanks
![image](https://user-images.githubusercontent.com/58068359/69542631-2a6f5980-0f8c-11ea-9255-a6ad988b630f.png)

Haydar

stephenitpro:
I just upgraded to 18.0.4 from 18.0.3, now I have this issue too.  In some cases I have triplicate group names.  Why????? I tried to delete one and it gave an error "An error occured during the request. Unable to proceed."

Running PHP 7.4

PVince81:
also see https://github.com/nextcloud/groupfolders/issues/1246 with more details about the symptoms: the group folders mount points are overlapping each other

PVince81:
ok, seems we already had an older issue about this, so closing as duplicate of https://github.com/nextcloud/groupfolders/issues/1246

fschrempf:
Reopening because issue is still existing in current master.

fschrempf:
According to https://github.com/nextcloud/groupfolders/issues/1721#issuecomment-965600129 it is a valid use case to have multiple groupfolders with the same name. This means we have to find a way to handle the collision for users that have access to multiple groupfolders with the same name.

lethargosapatheia:
I for one doesn't really understand how that is a valid case, also given the fact that basically everyone assumes that there shouldn't be any duplicates in the group folders.

Having group folders with the same name for a different set of users sounds to me like a multitenant setup, which you wouldn't want to have in the first place within a single nextcloud instance. You could simply run multiple instances for various clients if that's what you need and it would be much clear at that.
People don't expect to have multiple identical group folders just like they don't expect having folders on their computer with the same name in the same location.

[...]

PR #3752, opened 2025-04-24
fix: prevent duplicate mount points

skjnldsv:
Fix https://github.com/nextcloud-gmbh/customer-feature-requests/issues/1012
Partially addresses https://github.com/nextcloud/groupfolders/issues/681

### DIFF
--- a/lib/Controller/FolderController.php
+++ b/lib/Controller/FolderController.php
@@ -19,6 +19,7 @@ use OCP\AppFramework\Http\Attribute\FrontpageRoute;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\Attribute\PasswordConfirmationRequired;
 use OCP\AppFramework\Http\DataResponse;
+use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\AppFramework\OCS\OCSNotFoundException;
 use OCP\AppFramework\OCSController;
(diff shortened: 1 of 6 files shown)


icewind1991:
Do we actually want to do this? there are legitimate use cases for it and blocking it now might break existing workflows

icewind1991 (approved):
I'm predicting the need to add a config flag for this eventually, but for the majority of users this should be an improvement.

lib/Controller/FolderController.php
```diff
@@ -164,12 +165,28 @@ public function getFolder(int $id): DataResponse {
...
+
+		return null;
+	}
+
+	private function checkMountPointExists(string $mountpoint): ?DataResponse {
+		$storageId = $this->getRootFolderStorageId();
+		if ($storageId === null) {
+			throw new OCSNotFoundException('Groupfolder not found');
+		}
+
```
icewind1991, line 184: why is this here?

skjnldsv: Because psalm isn't smart enough to understand the check above :facepalm:

provokateurin: Then you should rather suppress it than bending your code for psalm to be happy.

skjnldsv: With something like `@var` ?

provokateurin: Yeah or just `@psalm-supress`

skjnldsv: Not sure if really cleaner, but here it is: https://github.com/nextcloud/groupfolders/pull/3802

PR #3808, opened 2025-05-26
Add pagination for admin interface

provokateurin:


provokateurin:
@susnux can you please pull all occurrences of the number of items per page into a constant so we can easily adjust it later instead of having to change 20 lines and potentially miss some.

susnux:
@provokateurin done

PR #3837, opened 2025-06-10
fix: add `orderby` to pagination to prevent sorting issues

susnux:
To reproduce:
- first create 51 folders like x1, x2... x51
- then after that create one folder like a1
- load page
  - see x1 - x50 are listed
  - go to page 2
  - see x50 and x51 are listed
  - go back to page 1
  - see that a1 is now listed

### DIFF
--- a/lib/Controller/FolderController.php
+++ b/lib/Controller/FolderController.php
@@ -40,6 +40,13 @@ use OCP\IUserSession;
 class FolderController extends OCSController {
 	private readonly ?IUser $user;
 
+	protected const ALLOWED_ORDER_BY = [
+		'mount_point',
+		'quota',
+		'groups',
+		'acl',
+	];
+
 	public function __construct(
 		string $AppName,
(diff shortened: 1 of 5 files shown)


provokateurin:
ksort is working, but results in an "unnatural" sorting order, right? I didn't really get that from the PR title and description as it doesn't really explain what is wrong.

susnux:
its currently sorted by ID but the frontend is sorted by displayname by default (mount point).
So we load the wrong batch per pagination, as the backend needs to sort the same way for the batch as the frontend does.

susnux:
BTW no idea whats wrong with psalm, it complains about possible wrong types, but its not possible.
The types it could have are only string, bool or int which all can be cast to string.

skjnldsv (approved):

provokateurin (approved):

provokateurin:
Fixed it

PR #3849, opened 2025-06-19
Allow per-groupfolder storage

icewind1991:
Removes the assumption that all groupfolders are on the same storage. Instead we track the storage id and root id with the folder.

Currently all groupfolders will still be on the root storage, but this opens up the option to have them on different storages down the line.

Included is a cleanup of moving the folder data from an array to a proper type to make it fully clear which api return what fields.


lib/ACL/ACLManager.php
```diff
@@ -94,13 +105,17 @@ private function getRelevantPaths(string $path): array {
 			}
 			[, , $groupFolderId, $rootTrashedItemName] = $parts;
 			$groupFolderId = (int)$groupFolderId;
-			/* Remove the date part */
-			$separatorPos = strrpos($rootTrashedItemName, '.d');
-			if ($separatorPos === false) {
-				throw new RuntimeException('Invalid trash item name ' . $rootTrashedItemName);
+			if (!$rootTrashedItemName) {
+				$fromTrashbin = false;
```
come-nc: What?

lib/AppInfo/Application.php
```diff
@@ -225,7 +225,6 @@ public function register(IRegistrationContext $context): void {
 				$c->get(IAppConfig::class),
 				$c->get(LoggerInterface::class),
 				$c->get(IUserMappingManager::class),
-				$rootFolderProvider
```
come-nc, line 228: You should remove the var definition as well.
Also, you can remove the whole registerService call I think, DI should be able to build the factory without any help.

lib/Folder/Folder.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
```
come-nc: strict_type missing

lib/Folder/Folder.php
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * SPDX-FileCopyrightText: 2025 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\GroupFolders\Folder;
+
+class Folder {
```
come-nc: I like having a class for this, but reusing the `Folder` name is confusing with OCP Folder class. Could it be FolderDefinition or something?

lib/Migration/Version20000Date20250612140256.php
```diff
@@ -0,0 +1,130 @@
...
+				$table->addColumn('root_id', Types::BIGINT, ['notnull' => false]);
+			}
+			if (!$table->hasColumn('storage_id')) {
+				$table->addColumn('storage_id', Types::BIGINT, ['notnull' => false]);
+			}
+			if (!$table->hasColumn('options')) {
+				$table->addColumn('options', Types::TEXT, ['notnull' => false]);
+			}
+		}
+		return $schema;
```
come-nc: only return schema if modifications are needed, otherwise return null

lib/ACL/ACLManager.php
icewind1991: Hmm, I was running into issues during tests where `$rootTrashedItemName` was empty, but that doesn't seem to be a problem anymore.

come-nc (approved):

PR #3859, opened 2025-06-23
build(composer): Stop updating vendor-bin when updating main composer.json

provokateurin:
https://github.com/nextcloud/groupfolders/blob/18fcd3eb9bf10ac5a4affaf9fae232cb5483dcc5/.github/workflows/lint-php-cs.yml#L48 causes a `composer update` in vendor-bin, which breaks CI because a new nextcloud/coding-standard version is pulled. This was already addressed in https://github.com/nextcloud/.github/pull/538, but removing this is the better option to prevent it from happening on manual composer modifications.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,9 +12,6 @@
 		"post-install-cmd": [
 			"@composer bin all install --ansi"
 		],
-		"post-update-cmd": [
-			"@composer bin all update --ansi"
-		],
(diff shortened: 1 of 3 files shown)


provokateurin:
Bumped psalm due to https://github.com/vimeo/psalm/issues/11213

PR #3867, opened 2025-06-24
fix(ACL): Apply folder permissions mask when testing path permission

provokateurin:
If a user has no access to a Team folder, no rules are returned and it is assumed that this user has all permissions.
Instead of applying the folder permissions as a mask, it would also be possible to just check if the folder permissions are 0 and then use 0 as the path permissions directly, but in the end this results in the same output.

### DIFF
--- a/lib/Command/ACL.php
+++ b/lib/Command/ACL.php
@@ -77,7 +77,7 @@ class ACL extends FolderCommand {
 				$path = $input->getArgument('path');
 				$aclManager = $this->aclManagerFactory->getACLManager($user);
 				$permissions = $aclManager->getACLPermissionsForPath($jailPath . rtrim('/' . $path, '/'));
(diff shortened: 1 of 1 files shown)


icewind1991:
This change also impacts users that do have access to the groupfolder, but the group they have access trough doesn't have full (non-acl) permissions.

For example, if a user has access trough a group that doesn't have the "delete" base permissions, and a `-share` acl on a folder.

With master the acl test returns

```
+read, +write, +create, +delete, -share
```

With this PR it returns

```
+read, +write, +create, -share
```

I'm not sure if either of those are "correct" output.

```
+read, +write, +create, -delete, -share
```

Is probably most correct (though it might mislead admins into thinking the `-delete` comes from an ACL rule, having an "explain" option that show how the permissions are calculated has been on my whish list for a while).

I would suggest either changing this PR to show that output, or scope it down to only handle the case where a user doesn't have access at all.

provokateurin:
@icewind1991 please review again.

PR #3887, opened 2025-07-11
remove trashbin layout assumptions from ACL handling

icewind1991:
The current ACL code has explicit logic for dealing with trashbin paths, encoding assumptions about how the groupfolders trash is laid out.

This reworks how ACLs are checked in the trashbin to remove these assumptions and allow for changing how the trash is stored in the future.

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -9,6 +9,7 @@ declare (strict_types=1);
 namespace OCA\GroupFolders\Trash;
 
 use OC\Encryption\Exceptions\DecryptionFailedException;
+use OC\Files\ObjectStore\ObjectStoreStorage;
 use OC\Files\Storage\Wrapper\Encryption;
 use OC\Files\Storage\Wrapper\Jail;
 use OCA\Files_Trashbin\Expiration;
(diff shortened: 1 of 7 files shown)


lib/Trash/TrashBackend.php
```diff
@@ -332,6 +335,25 @@ private function userHasAccessToFolder(IUser $user, int $folderId): bool {
...
+		GroupTrashItem $item,
+		int $permission = Constants::PERMISSION_READ,
+		string $pathInsideItem = '',
+	): bool {
+		try {
+			$aclManager = $this->aclManagerFactory->getACLManager($item->getUser());
+			$trashPath = $this->getUnJailedPath($item->getTrashNode()) . $pathInsideItem;
+			$activePermissions = $aclManager->getACLPermissionsForPath($item->getGroupFolderStorageId(), $trashPath, '__groupfolders/trash');
+			$originalPath = $item->folder->rootCacheEntry->getPath() . '/' . $item->getInternalOriginalLocation() . $pathInsideItem;
+			$originalLocationPermissions = $aclManager->getACLPermissionsForPath($item->getGroupFolderStorageId(), $originalPath);
```
come-nc, line 348: Would be easier to read if we added a method `getOriginalPath` on `GroupThrashItem`.

come-nc: Wait, but there is `getOriginalLocation` on trash item. It’s not the same thing as $originalPath here?

come-nc (approved):
This is really complex. I’m gonna trust the e2e tests which were added last time I suppose.
Copie d'écran_20250825_110641

It seems there is no test for delete, and then delete parent. As there is code related to this situation it may be good to test it.

icewind1991:
> It seems there is no test for delete, and then delete parent

There is in the unit tests

PR #3917, opened 2025-08-11
[31] store the fileid of the root of groupfolders

icewind1991:
Partial backport of https://github.com/nextcloud/groupfolders/pull/3849 + https://github.com/nextcloud/groupfolders/pull/3886

This doesn't attempt to allow for multiple storages to be used, but uses the storage fileid of the groupfolder root to optimize getting the file metadata for groupfolders, as it can now simply join by fileid instead of having to do a parent+name.

### DIFF
--- a/lib/Mount/FolderStorageManager.php
+++ b/lib/Mount/FolderStorageManager.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2025 Robin Appelman <robin@icewind.nl>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\GroupFolders\Mount;
+
+use OC\Files\Storage\Wrapper\Jail;
+use OCP\Files\Folder;
+use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
+use OCP\Files\Storage\IStorage;
+use OCP\IAppConfig;
+
+class FolderStorageManager {
+	private readonly bool $enableEncryption;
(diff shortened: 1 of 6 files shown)


lib/Mount/FolderStorageManager.php
```diff
@@ -0,0 +1,79 @@
...
+
+class FolderStorageManager {
+	private readonly bool $allowRootShare;
+	private readonly bool $enableEncryption;
+
+	public function __construct(
+		private readonly IRootFolder $rootFolder,
+		private readonly IAppConfig $appConfig,
+	) {
+		$this->allowRootShare = $this->appConfig->getValueString('groupfolders', 'allow_root_share', 'true') === 'true';
```
come-nc: Why is that not getValueBool?

icewind1991: Copy-paste from old code, changed

lib/Mount/FolderStorageManager.php
```diff
@@ -0,0 +1,79 @@
...
+
+class FolderStorageManager {
+	private readonly bool $allowRootShare;
+	private readonly bool $enableEncryption;
+
+	public function __construct(
+		private readonly IRootFolder $rootFolder,
+		private readonly IAppConfig $appConfig,
+	) {
+		$this->allowRootShare = $this->appConfig->getValueBool('groupfolders', 'allow_root_share', true);
```
come-nc: This one is unused?

icewind1991: Fixed

come-nc (approved):

PR #3920, opened 2025-08-19
per groupfolder storage

icewind1991:
- [x] https://github.com/nextcloud/groupfolders/pull/3849
- [x] https://github.com/nextcloud/groupfolders/pull/3887
- [x] https://github.com/nextcloud/groupfolders/pull/3941
- [x] https://github.com/nextcloud/groupfolders/pull/3921
- [x] https://github.com/nextcloud/groupfolders/pull/3943
- [x] actually setup different storages per groupfolder (this PR)

Currently groupfolders are stored under `__groupfolders` in the root storage, using jail mounts to mount the individual groupfolders for the users. This was done to simplify the app's code (no need for special object store handling etc) but comes with the limitation that, in multi-bucket object store setups, all groupfolder files are still stored on the same bucket.

This changes things to instead create a single storage per groupfolder (jails are still being used to separate out the files, trash and versions as separate mounts, while keeping those stored together logically).

For local primary storage, the files are now stored in `$datadir/__groupfolders/$folderId/files` (and `$datadir/__groupfolders/$folderId/trash`/`$datadir/__groupfolders/$folderId/versions` for trashbin/versions) where `$datadir/__groupfolders/$folderId` is the root of the per-groupfolder local storage.

For object store, similar logic to home storage is used to support multi-instance and multi-bucket object store but with folder ids instead of user ids.

All the new logic only affects newly created groupfolders, existing ones keep the old storage layout

### DIFF
--- a/lib/Mount/FolderStorageManager.php
+++ b/lib/Mount/FolderStorageManager.php
@@ -41,8 +41,8 @@ class FolderStorageManager {
 	/**
 	 * @return array{storage_id: int, root_id: int}
 	 */
-	public function getRootAndStorageIdForFolder(int $folderId, bool $separateStorage): array {
-		$storage = $this->getBaseStorageForFolder($folderId, $separateStorage);
+	public function initRootAndStorageForFolder(int $folderId, bool $separateStorage): array {
+		$storage = $this->getBaseStorageForFolder($folderId, $separateStorage, init: true);
 		$cache = $storage->getCache();
 		$id = $cache->getId('');
 		if ($id === -1) {
@@ -68,9 +68,10 @@ class FolderStorageManager {
 		?IUser $user = null,
 		bool $inShare = false,
 		string $type = 'files',
+		bool $init = false,
 	): IStorage {
 		if ($separateStorage) {
-			return $this->getBaseStorageForFolderSeparate($folderId, $folder, $user, $inShare, $type);
+			return $this->getBaseStorageForFolderSeparate($folderId, $folder, $user, $inShare, $type, $init);
 		} else {
 			return $this->getBaseStorageForFolderRootJail($folderId, $folder, $user, $inShare, $type);
 		}
@@ -85,11 +86,12 @@ class FolderStorageManager {
 		?IUser $user = null,
 		bool $inShare = false,
 		string $type = 'files',
+		bool $init = false,
 	): IStorage {
 		if ($this->primaryObjectStoreConfig->hasObjectStore()) {
-			$storage = $this->getBaseStorageForFolderSeparateStorageObject($folderId);
+			$storage = $this->getBaseStorageForFolderSeparateStorageObject($folderId, $init);
 		} else {
-			$storage = $this->getBaseStorageForFolderSeparateStorageLocal($folderId);
+			$storage = $this->getBaseStorageForFolderSeparateStorageLocal($folderId, $init);
 		}
 
 		if ($folder?->acl && $user) {
@@ -117,15 +119,18 @@ class FolderStorageManager {
 
 	private function getBaseStorageForFolderSeparateStorageLocal(
 		int $folderId,
+		bool $init = false,
 	): IStorage {
 		$dataDirectory = $this->config->getSystemValue('datadirectory');
 		$rootPath = $dataDirectory . '/__groupfolders/' . $folderId;
-		$init = !is_dir($rootPath);
 		if ($init) {
-			mkdir($rootPath, 0777, true);
-			mkdir($rootPath . '/files');
-			mkdir($rootPath . '/trash');
-			mkdir($rootPath . '/versions');
+			$result = mkdir($rootPath . '/files', recursive:  true);
+			$result = $result && mkdir($rootPath . '/trash');
+			$result = $result && mkdir($rootPath . '/versions');
+
+			if (!$result) {
+				throw new \Exception('Failed to create base directories for group folder ' . $folderId);
+			}
 		}
 
 		$storage = new Local([
@@ -140,6 +145,7 @@ class FolderStorageManager {
 
 	private function getBaseStorageForFolderSeparateStorageObject(
 		int $folderId,
+		bool $init = false,
 	): IStorage {
 		$objectStoreConfig = $this->primaryObjectStoreConfig->getObjectStoreConfiguration($this->getObjectStorageKey($folderId));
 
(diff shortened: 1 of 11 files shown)


lib/Mount/FolderStorageManager.php
```diff
@@ -54,7 +60,118 @@ public function getRootAndStorageIdForFolder(int $folderId): array {
...
+		bool $inShare = false,
+		string $type = 'files',
+	): IStorage {
+		if ($this->primaryObjectStoreConfig->hasObjectStore()) {
+			$storage = $this->getBaseStorageForFolderSeparateStorageObject($folderId);
+		} else {
+			$storage = $this->getBaseStorageForFolderSeparateStorageLocal($folderId);
+		}
+
+		if ($folder && $folder->acl && $user) {
```
provokateurin:
```suggestion
		if (($folder?->acl ?? false) && $user !== null) {
```

lib/Mount/FolderStorageManager.php
```diff
@@ -54,7 +57,72 @@ public function getRootAndStorageIdForFolder(int $folderId): array {
...
+		?FolderDefinition $folder = null,
+		?IUser $user = null,
+		bool $inShare = false,
+		string $type = 'files',
+	): IStorage {
+		$dataDirectory = $this->config->getSystemValue('datadirectory');
+		$rootPath = $dataDirectory . '/__groupfolders/' . $folderId;
+		$init = !is_dir($rootPath);
+		if ($init) {
+			mkdir($rootPath, 0777, true);
```
provokateurin: 777 permission doesn't seem like a good idea? Should be 750 or something like that

lib/Mount/FolderStorageManager.php
```diff
@@ -54,7 +57,72 @@ public function getRootAndStorageIdForFolder(int $folderId): array {
...
+			mkdir($rootPath . '/versions');
+		}
+
+		$storage = new Local([
+			'datadir' => $rootPath,
+		]);
+
+		if ($init) {
+			$storage->getScanner()->scan('');
+		}
```
provokateurin: Why not use `$storage->mkdir()` to create these folders?

lib/Mount/FolderStorageManager.php
provokateurin: (Same for the first commit, but I couldn't suggest a change there)

lib/Mount/FolderStorageManager.php
```diff
@@ -99,24 +134,38 @@ public function getBaseStorageForFolderSeparateStorageLocal(
...
+		]);
+		$arguments['storageid'] = 'object::groupfolder:' . $folderId . '.' . $objectStore->getStorageId();
+
+		$storage = new ObjectStoreStorage($arguments);
+
+		if (!$storage->file_exists('files')) {
+			$storage->mkdir('files');
+			$storage->mkdir('trash');
+			$storage->mkdir('versions');
 		}
```
provokateurin: Each of the folders should be checked for existence. If creating `files` works bu the other two don't, then it will run into problems later as the directories are not recreated.

lib/Mount/FolderStorageManager.php
```diff
@@ -106,11 +223,55 @@ public function getBaseStorageForFolder(int $folderId, ?FolderDefinition $folder
...
 			$cache->clear();
 			$storage->rmdir('');
 		}
 	}
+
+	private function getObjectStorageKey(int $folderId): string {
+		$configs = $this->primaryObjectStoreConfig->getObjectStoreConfigs();
+		if ($this->primaryObjectStoreConfig->hasMultipleObjectStorages()) {
+			$configKey = 'object_store_key_' . $folderId;
+			$storageConfigKey = $this->appConfig->getValueString('groupfolders', $configKey);
```
provokateurin: Please use \OCA\GroupFolders\AppInfo\Application::APP_ID everywhere.

lib/Mount/FolderStorageManager.php
```diff
@@ -106,11 +223,55 @@ public function getBaseStorageForFolder(int $folderId, ?FolderDefinition $folder
...
+
+	// logic taken from OC\Files\ObjectStore\Mapper which we can't use because it requires an IUser
+	private function calculateBucketNum(string $key, array $objectStoreConfig): string {
+		$numBuckets = $objectStoreConfig['arguments']['num_buckets'] ?? 64;
+
+		// Get the bucket config and shift if provided.
+		// Allow us to prevent writing in old filled buckets
+		$minBucket = isset($objectStoreConfig['arguments']['min_bucket'])
+			? (int)$objectStoreConfig['arguments']['min_bucket']
+			: 0;
```
provokateurin:
```suggestion
		$minBucket = $objectStoreConfig['arguments']['min_bucket'] ?: 0;
```
I think this should also work :thinking:

provokateurin (changes requested):

lib/Mount/FolderStorageManager.php
icewind1991: `0777` is the default argument, the folder won't actually be `0777` because of the umask though

come-nc:
```suggestion
			mkdir($rootPath, recursive: true);
```
This would be clearer then.

icewind1991: what is this new fangled sorcery

lib/Mount/FolderStorageManager.php
icewind1991: reworked the initialization to be more explicit and fail loudly.

lib/Mount/FolderStorageManager.php
icewind1991: I want to do "raw" `mkdir` anyway because of the recursive, so might as well do that for the rest, `$storage->mkdir` has no advantage over raw mkdir here.

lib/Mount/FolderStorageManager.php
```diff
@@ -54,7 +61,128 @@ public function getRootAndStorageIdForFolder(int $folderId): array {
...
+	private function getBaseStorageForFolderSeparateStorageLocal(
+		int $folderId,
+		bool $init = false,
+	): IStorage {
+		$dataDirectory = $this->config->getSystemValue('datadirectory');
+		$rootPath = $dataDirectory . '/__groupfolders/' . $folderId;
+		if ($init) {
+			$result = mkdir($rootPath . '/files', recursive:  true);
+			$result = $result || mkdir($rootPath . '/trash');
+			$result = $result || mkdir($rootPath . '/versions');
```
come-nc: With the `||` I believe PHP will not run the command if result is already true.
Should be `&&` I believe?

icewind1991: good catch

PR #3921, opened 2025-08-19
fix: make version backend storage location agnostic

icewind1991:
Removes assumptions from how groupfolders are stored from the versions backend, instead always go trough a version mount.

In preparation of supporting per-groupfolder storages.

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -11,6 +11,7 @@ namespace OCA\GroupFolders\Mount;
 use OC\Files\Storage\Wrapper\PermissionsMask;
 use OCA\GroupFolders\ACL\ACLManager;
 use OCA\GroupFolders\ACL\ACLManagerFactory;
+use OCA\GroupFolders\Folder\FolderDefinition;
 use OCA\GroupFolders\Folder\FolderDefinitionWithPermissions;
 use OCA\GroupFolders\Folder\FolderManager;
 use OCP\Constants;
@@ -90,7 +91,7 @@ class MountProvider implements IMountProvider {
 				$loader,
 				$user,
 				$aclManager,
-				$rootRules
+				$rootRules,
 			);
 		}, $folders));
 	}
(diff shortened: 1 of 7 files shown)


lib/Mount/MountProvider.php
```diff
@@ -168,18 +169,51 @@ public function getTrashMount(
...
+			if (!$cacheEntry) {
+				$storage->getScanner()->scan('');
+				$cacheEntry = $storage->getCache()->get('');
+				if (!$cacheEntry) {
+					throw new \Exception('Group folder version root is not in cache even after scanning for folder ' . $folder->id);
+				}
+			}
+		}
+
+		$trashStorage = $this->getGroupFolderStorage(
```
come-nc:
```suggestion
		$versionsStorage = $this->getGroupFolderStorage(
```

lib/Versions/GroupVersionsExpireManager.php
```diff
@@ -8,7 +8,6 @@
 
 namespace OCA\GroupFolders\Versions;
 
-use OC\Files\View;
```
come-nc, line 11: :tada:

lib/Versions/VersionsBackend.php
```diff
@@ -275,27 +281,27 @@ public function getAllVersionedFiles(FolderDefinitionWithMappings $folder): arra
...
+		$mountPoint = '/dummy/files_versions/groupfolders/' . $folder->id;
+		$mount = $this->mountManager->find($mountPoint);
+		if ($mount->getMountPoint() !== $mountPoint) {
+			$trashMount = $this->mountProvider->getVersionsMount(
+				$folder,
+				$mountPoint,
+				$this->storageFactory,
+			);
+			$this->mountManager->addMount($trashMount);
 		}
```
come-nc: Is this named trashMount by copy/paste as well?
I’d welcome a comment here, because I do not understand what `$mount->getMountPoint() !== $mountPoint` implies.

come-nc: Also NotFoundException from calling `find` should be caught, no?

icewind1991: Added a comment,

`find` throwing a `NotFoundException` means that the root storage isn't setup so things are very broken and the error should be propagated.

CarlSchwan (approved):

come-nc (approved):

PR #3953, opened 2025-09-03
fix: ensure user id is string in search result

icewind1991:
Ensures that numeric user ids can be assigned as ACL manager


provokateurin (approved):

provokateurin:
Sounds like `displayNamesInGroup` in server is doing something wrong though, maybe you can fix it there as well @icewind1991 ?

icewind1991:
> Sounds like `displayNamesInGroup` in server is doing something wrong though, maybe you can fix it there as well @icewind1991 ?

Unfortunately, it's just not possible to use a numeric string as an array key and not have php automatically cast it: https://3v4l.org/HR3tF

Issue #3972, opened 2025-09-10
Groupfolders are shown empty after upgrading to 18.1.5 / 19.1.4


oleua:
Hi!

I've updated the app from 18.1.4 to 18.1.5 and now all my groupfolders are shown empty, though the files are present in the directories.

In logs I see constantly:
```
[...]

PR #3978, opened 2025-09-12
fix: more reliably determine root storage id in migration

icewind1991:
This should hopefully resolve the migration issues that seem to be causing https://github.com/nextcloud/groupfolders/issues/3972


icewind1991:
renamed to migration to ensure it is ran again for people who ran it previously and reverted.

icewind1991:
Made the updated migration also fix incorrectly migrated values

PR #4071, opened 2025-10-13
chore: Nextcloud 33 is PHP 8.2+

susnux:

PR #4074, opened 2025-10-14
fix: Fix sidebar not loading

provokateurin:
The sidebar is not loading anymore since https://github.com/nextcloud/server/pull/55524, because `OC.MimeType._mimeTypeIcons` not longer exists.
The icon is actually no longer handled by the app itself, but by https://github.com/nextcloud/server/blob/ffbbdb2d04c900bcc4ed6bbcf97d4e5fc30dfad9/apps/files/src/components/FileEntry/FileEntryPreview.vue#L236, so it can simply be removed.


provokateurin:
Psalm is fixed by https://github.com/nextcloud/groupfolders/pull/4071, but since this is a frontend only change it is safe to ignore.

Issue #4073, opened 2025-10-13
Performance regression in after 20.1.0


SystemKeeper:
### Steps to reproduce
Noticed while working on talk

1.  Have a plain conversation without files and check the query count for the `/context` endpoint -> 14 queries
2. Share a vCard to that conversation
3. Check query count again for `/context` endpoint -> **1263** queries 💣 

Number of groupfolders on the system:

```bash
occ groupfolders:list | grep -v '+' | wc -l
     608
```

Queries look like this:

```sql
CG1qhSP33yOcmsxJTwVd	 primary  0 SELECT `filecache`.`fileid`, `storage`, `path`, `path_hash`, `filecache`.`parent`, `filecache`.`name`, `mimetype`, `mimepart`, `size`, `mtime`, `storage_mtime`, `encrypted`, `filecache`.`etag`, `filecache`.`permissions`, `checksum`, `unencrypted_size`, `metadata_etag`, `creation_time`, `upload_time`, `meta`.`json` AS `meta_json`, `meta`.`sync_token` AS `meta_sync_token` FROM `oc_filecache` `filecache` LEFT JOIN `oc_filecache_extended` `fe` ON `filecache`.`fileid` = `fe`.`fileid` LEFT JOIN `oc_files_metadata` `meta` ON `filecache`.`fileid` = `meta`.`file_id` WHERE (`path_hash` = :dcValue1) AND (`storage` = :dcValue2); {"dcValue1":"29ff0edf73a32cb03e437d88fd049245","dcValue2":1};
[...]

PR #4075, opened 2025-10-14
perf(FolderStorageManager): Cache root groupfolders

provokateurin:
Fixes https://github.com/nextcloud/groupfolders/issues/4073

Additionally I also cached the subfolders for trash and versions, as they are the same across all groupfolders.

(This only affects groupfolders created before stable32, so make sure to test accordingly).

### DIFF
--- a/lib/Mount/FolderStorageManager.php
+++ b/lib/Mount/FolderStorageManager.php
@@ -27,6 +27,7 @@ use OCP\IUser;
 
 class FolderStorageManager {
 	private readonly bool $enableEncryption;
+	private array $cachedFolders = [];
 
 	public function __construct(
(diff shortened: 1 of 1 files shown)


SystemKeeper:
> (This only affects groupfolders created before stable32, so make sure to test accordingly).

Not sure, is the difference not separate vs. non-separate storage for groupfolders?

provokateurin:
Yes it is, see the codepath where this method is called.

SystemKeeper:
Okay, but then it also affects new groupfolders, just without separate storage?

provokateurin:
No any new groupfolders will have a separate storage.

PR #4096, opened 2025-10-17
fix: numbers and underscore are not sorted correctly

leftybournes:
This fixes the issue of team folders not being sorted properly when there are more than one page of folders (>50 folders), especially when a folder name starts with an underscore. When a folder name starts with an underscore, it disappears until the last page is revealed and then suddenly appears in the first page. This also naturally sorts folder names with numbers.

### DIFF
--- a/lib/Controller/FolderController.php
+++ b/lib/Controller/FolderController.php
@@ -150,13 +150,13 @@ class FolderController extends OCSController {
 					return $value;
 				}
 			} else {
-				if (($value = strcmp((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
+				if (($value = $this->compareFolderNames((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
 					return $value;
 				}
 			}
 
 			// fallback to mount_point
-			if (($value = strcmp($a['mount_point'] ?? '', $b['mount_point'])) !== 0) {
+			if (($value = $this->compareFolderNames($a['mount_point'] ?? '', $b['mount_point'])) !== 0) {
 				return $value;
 			}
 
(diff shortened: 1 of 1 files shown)


lib/Controller/FolderController.php
```diff
@@ -150,13 +150,30 @@ public function getFolders(bool $applicable = false, int $offset = 0, ?int $limi
 					return $value;
 				}
 			} else {
-				if (($value = strcmp((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
+				if (($value = strnatcmp((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
+					// Folder names starting with '_' get pushed to the end while they are brought to the
+					// beginning in the frontend. Do the same here to keep it consistent with the frontend
+					if (strpos($a[$orderBy], '_') === 0 && strpos($b['mount_point'], '_') !== 0) {
```
salmart-dev: Is the comparison with $b's mount_point, instead of $orderBy, intentional here and in the next if?

leftybournes: No, it's a mistake. Thanks for the catch.

lib/Controller/FolderController.php
```diff
@@ -150,13 +150,30 @@ public function getFolders(bool $applicable = false, int $offset = 0, ?int $limi
...
 			} else {
-				if (($value = strcmp((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
+				if (($value = strnatcmp((string)($a[$orderBy] ?? ''), (string)($b[$orderBy] ?? ''))) !== 0) {
+					// Folder names starting with '_' get pushed to the end while they are brought to the
+					// beginning in the frontend. Do the same here to keep it consistent with the frontend
+					if (strpos($a[$orderBy], '_') === 0 && strpos($b[$orderBy], '_') !== 0) {
+						return -1;
+					}
+
+					if (strpos($b[$orderBy], '_') === 0 && strpos($a[$orderBy], '_') !== 0) {
```
salmart-dev: To avoid scanning strings:

```suggestion
					if (strncmp($a[$orderBy], '_', 1) === 0 && strncmp( $b[$orderBy], '_', 1 ) !== 0) {
						return -1;
					}

					if (strncmp($b[$orderBy], '_', 1) === 0 && strncmp( $a[$orderBy], '_', 1 ) !== 0) {
```

salmart-dev (approved):
Tested and the current fix works. I added another comment for an optimization to avoid scanning the names of folders up to 8 times just to find out if they have an _ at the beginning or not.

artonge:
Should have seen it coming, you need to update the cypress version about files_versions. You can copy past what I did in https://github.com/nextcloud/server/pull/55965/files#diff-2624570d4890032e1950b55bc6ab8d8e4bbc2e5d9f86f200254e05c7d64863a5

leftybournes:
> Should have seen it coming, you need to update the cypress version about files_versions. You can copy past what I did in https://github.com/nextcloud/server/pull/55965/files#diff-2624570d4890032e1950b55bc6ab8d8e4bbc2e5d9f86f200254e05c7d64863a5

Thank you for the help. Here's the PR that does the same. There's progress but it's still failing.

https://github.com/nextcloud/groupfolders/pull/4110

PR #4110, opened 2025-10-28
fix: cypress tests

leftybournes:


artonge:
They were broken due to the `files_versions` migration to vue3: https://github.com/nextcloud/server/pull/55965

Sorry for that.

Issue #2779, opened 2024-01-23
Have a virtual group "all" for the advanced permissions 


schiessle:
When using group folders it happens to me regularly I want to remove from everyone the write permissions, e.g. that they can't accidentally break my documents.

What I'm doing at the moment:

1. I remove for every group with access to the group folder the "write" and "delete" permissions
2. I add for myself all permissions again

This has some obvious shoirtcummings:

1. If the admin give later additional groups access to the folder they are not covered and can edit my documents
2. If a lot of groups have access to the folder it is quite some work to remove the rights from each group

**Proposed solution:**

The "advanced permissions" drop-down has a virtual group called "all users". This way I can modify the access rights for everyone with one click.

anoymouserver:
I've never used it, but you might find the apps [Everyone Group](https://apps.nextcloud.com/apps/group_everyone) or [Auto Groups](https://apps.nextcloud.com/apps/auto_groups) useful as an alternative.

provokateurin:
@schiessle can you elaborate on shortcoming 1? I don't really understand the problem.
As for 2 we could improve the way you can change permissions, e.g. bulk changing by clicking on a column or shift-drag selection.

schiessle:
@provokateurin Thanks for looking into the issue!


The shortcoming is simple. You have a large instance, many users and groups. You have a shared group folder, let's say a company wide folder for all company related files. Now you want to  allow only a single person or group access to a subfolder. At the moment. You have to do the following:

1. Deny all groups access to the subfolder which also means that you need to know all groups with access to the group folder. This is already a impossible assumption, but if you know all groups this can be already a huge list of advanced permissions you have to create one by one
2. Than you have to grant the specific group/persons access rights again
3. Whenever a new group is introduced after you configured the advanced permissions they will have access to the subfolder. You and everyone else who manages advanced permissions would always need to remember all folders they configured this way, get to know that a new group was introduced and adjust the advanced permissions, hopefully really fast so that people don't get access to files they shouldn't access -> error prone and simply not possible.

This would be solved if the advanced permissions would know a virtual group "all users", where In can set the permissions for all users without knowing anything about user and group structure on the system and without having to worry that something changes to the user/group structure.

I hope this makes it clear.

provokateurin:
Ok I understand now, but I don't know if this is something we can fix this way.
We will need to take a closer look later when we actually try to implement this.

provokateurin:
[...]

PR #4126, opened 2025-11-04
feat(ACL): Support granting no permissions by default

provokateurin:
Closes https://github.com/nextcloud/groupfolders/issues/2779

TODO:
- [x] Add flag to OCC command to toggle the option
- [x] Check again that all `Constants::PERMISSION_ALL` occurrences are correct and shouldn't be switched to the configurable base permission
- [x] Better frontend


provokateurin:
The frontend part probably needs to look a bit nicer, but I'll work on that. Everything else is ready to review.

lib/Controller/FolderController.php
```diff
@@ -601,4 +602,28 @@ private function compareFolderNames(string $a, string $b): int {
 
 		return $value;
 	}
+
+	/**
+	 * Toggle the ACL inverted base permission for a team folder
+	 *
+	 * @param int $id ID of the team folder
+	 * @param bool $acl_inverted_base_permission Whether ACL should be enabled or not
```
come-nc: description is incorrect

provokateurin:
Frontend needs some larger rework, but backend should be fine now.

provokateurin:
Frontend has been fixed:
1. The checkbox is only shown on creation, you never want to toggle it (unless we add a proper migration path)
2. The inherited permissions are always shown in the sidebar, which is an improvement over the current state, regardless of this feature.

lib/Migration/Version2100000Date20251104151854.php
```diff
@@ -0,0 +1,39 @@
...
+	#[Override]
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('group_folders');
+		$table->addColumn('acl_default_no_permission', Types::BOOLEAN, [
+			'notnull' => true,
+			'default' => false,
+		]);
```
come-nc: It should check if the column is missing first

provokateurin:
Something went wrong here, I must have force pushed an earlier version again.

lib/DAV/ACLPlugin.php
```diff
@@ -104,33 +101,37 @@ public function propFind(PropFind $propFind, INode $node): void {
 		}
 
 		$propFind->handle(self::ACL_LIST, function () use ($fileInfo, $mount): ?array {
+			// Happens when sharing with a remote instance
```
joshtrichards, line 104: Any idea why we can't just have a single `$this->user === null` guard at the top of `propFind()`? I know the handlers can be lazy loaded, but in practice `$this->user` is immutable. I'm doing a refactor and moving `$this->user` to the constructor and making it readonly anyhow. `propPatch()` already has the guard at the top. I don't see any benefit here.

joshtrichards: I guess *some* of the handlers (the ones that don't depend on `$this->user` would then be prevented from registering too, but is that a practical real-world concern? The "Happens when sharing with a remote instance" makes me think there is some virtue in some of these handlers I guess still working and/or returning empty values (rather than being unavailable outright) under some circumstances?

provokateurin: That might work as well, I didn't check when refactoring this :sweat_smile:

PR #4153, opened 2025-11-17
chore: Run rector with php 8.2

CarlSchwan:


lib/Command/Scan.php
```diff
@@ -109,7 +109,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 				}
 			});
 
-			$scanner->listen('\OC\Files\Cache\Scanner', 'scanFolder', function (string $path) use ($output, &$statsRow): void {
+			$scanner->listen(\OC\Files\Cache\Scanner::class, 'scanFolder', function (string $path) use ($output, &$statsRow): void {
```
provokateurin: This is wrong. See https://github.com/nextcloud/groupfolders/pull/3850. I wanted to add a rector ignore, but didn't have the time yet. Maybe you can do that here?

CarlSchwan: sigh :/

CarlSchwan: I reverted that specific change for now

Issue #3708, opened 2025-03-19
Confusion between the "displayName" key and its actual value


zak39:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce

1. Go to the user settings page
2. Create a group (example: `Octopus`)
3. Rename this to `Fish`
4. Go to the groupfolders page
5. Create a groupfolder and add the created group previously
6. Open the web console (F12) and select network tab
7. Empty all results
8. Refresh your tab
9. Then, select the `folders` request

### Expected behaviourTell us what should happen

I should have an accurate state of my Groupfolders with the correct data.

### Actual behaviour

I get this response example : 

```bash
{
    "ocs": {
        "meta": {
            "status": "ok",
            "statuscode": 100,
            "message": "OK",
            "totalitems": "",
            "itemsperpage": ""
        },
        "data": {
            "1": {
                "id": 1,
                "mount_point": "Espace01",
                "groups": {
                    "SPACE-GE-1": 31,
                    "SPACE-U-1": 31,
                    "Poulpe-1": 31,
                    "Clown Fish-1": 31
                },
                "quota": -3,
                "size": 0,
                "acl": true,
                "manage": [
                    {
                        "type": "group",
                        "id": "SPACE-GE-1",
                        "displayname": "WM-Espace01"
                    }
                ],
                "group_details": {
                    "SPACE-GE-1": {
                        "displayName": "SPACE-GE-1",
                        "permissions": 31,
                        "type": "group"
                    },
                    "SPACE-U-1": {
                        "displayName": "SPACE-U-1",
                        "permissions": 31,
                        "type": "group"
                    },
                    "Poulpe-1": {
                        "displayName": "Poulpe-1",
                        "permissions": 31,
                        "type": "group"
                    },
                    "Clown Fish-1": {
                        "displayName": "Clown Fish-1",
                        "permissions": 31,
                        "type": "group"
                    }
                }
            },
            // ...
        }
    }
}
```

In the `group_details` key, the obtained displayname is the gid, not `display name`.

For example, here, the `display name` from `Clown Fish-1` printed here is `Clown Fish-1`.

However, the actual display name is `G-Clown Fish-Espace01` : 

[...]

PR #4162, opened 2025-11-19
fix(FolderManager): Use group display name instead of id for display name

provokateurin:
https://github.com/nextcloud/groupfolders/issues/3708

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -258,7 +258,7 @@ class FolderManager {
 						'displayname' => $user->getDisplayName(),
 					];
 				case 'group':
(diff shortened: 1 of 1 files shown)


lib/Folder/FolderManager.php
```diff
@@ -364,7 +364,7 @@ private function getAllApplicable(): array {
 				$entityId = (string)$row['group_id'];
 
 				$entry = [
-					'displayName' => $row['group_id'],
+					'displayName' => Server::get(IGroupManager::class)->get($row['group_id'])?->getDisplayName() ?? $row['group_id'],
```
artonge: Can't use DI?

provokateurin: Done

PR #4200, opened 2025-12-08
fix: Forbid deletion of current version in version expiration code

come-nc:
Log it instead.


come-nc:
@provokateurin Do you know if this cypress test is flaky? It fails and is related to versions so it’s scary.

provokateurin:
Yes this one is particularly flaky, but it would be great to actually have it pass. Maybe you can restart it later today, when CI has more headroom.

PR #4234, opened 2025-12-22
fix(dav): Register DAV plugins via SabrePluginAddEvent

joshtrichards:
A refactor that modernizes slightly, by switching SabreDAV plugin loading over to the event listener. Drops Sabre plugin management declarations from `appinfo/info.xml`.

Doesn't make any real difference at the moment, but less future code debt.

Misc:
- Adds new stubs needed for Psalm for this change
- Updates the static-code-analysis to explicitly use PHP 8.2 (same as Server) since this repo's Psalm isn't sufficiently up-to-date to support PHP 8.5 yet (tests were failing)

TODO:

- [x] ~~Split out Psalm php version range temporary lock into separate PR~~
- [x] ~~Drop Psalm php version lock once split out is merged~~
- [x] Squash/clean-up commits before merge

### DIFF
--- a/.github/workflows/psalm.yml
+++ b/.github/workflows/psalm.yml
@@ -38,7 +38,8 @@ jobs:
       - name: Set up php${{ steps.versions.outputs.php-available }}
         uses: shivammathur/setup-php@0f7f1d08e3e32076e51cae65eb0b0c871405b16e # v2.34.1
         with:
-          php-version: ${{ steps.versions.outputs.php-available }}
+          # php-version: ${{ steps.versions.outputs.php-available }}
+          php-version: '8.2'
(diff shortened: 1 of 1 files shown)


come-nc (commented):
I’m afraid about this because there are code paths which loads plugins from `PluginManager::getAppPlugins` which are not triggering `SabrePluginAddEvent`.
Also there is the ServerFactory using `OCP\SabrePluginEvent` :shrug: 

So I’d favor a cleanup of sabre server building on server side before migrating applications, unless you’re sure the plugins from groupfolders are only needed in the code paths using SabrePluginAddEvent ?

.github/workflows/psalm.yml
```diff
@@ -38,7 +38,8 @@ jobs:
       - name: Set up php${{ steps.versions.outputs.php-available }}
         uses: shivammathur/setup-php@0f7f1d08e3e32076e51cae65eb0b0c871405b16e # v2.34.1
         with:
-          php-version: ${{ steps.versions.outputs.php-available }}
+          # php-version: ${{ steps.versions.outputs.php-available }}
+          php-version: '8.2'
```
icewind1991, line 42: why is this changed?

PR #4238, opened 2025-12-31
refactor(MountProvider): Small optimizations

solracsf:


lib/Mount/MountProvider.php
```diff
@@ -268,7 +256,7 @@ private function findConflictsForUser(IUser $user, array $mountPoints): array {
 		$paths = [];
 		foreach (array_chunk($pathHashes, 1000) as $chunk) {
 			$query->setParameter('chunk', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
-			$paths = array_merge($paths, $query->executeQuery()->fetchAll(\PDO::FETCH_COLUMN));
+			array_push($paths, ...$query->executeQuery()->fetchAll(\PDO::FETCH_COLUMN));
```
come-nc, line 259: I’m afraid this may be less efficient, it will expand the whole array into function parameters instead of passing 2 arrays to array_merge. Do you have a reference on why this would be better?

solracsf: Each call to `array_merge()` creates a new array by copying all existing elements plus the new ones. `array_push()` with the spread operator `...` appends elements directly to the existing array without creating copies.

However, you might be right that spreading into function parameters isn't ideal.
The optimal approach may be to collect arrays first, then merge once:

```
$chunks = [];
foreach (array_chunk($pathHashes, 1000) as $chunk) {
	$query->setParameter('chunk', $chunk, IQueryBuilder::PARAM_STR_ARRAY);
	$chunks[] = $query->executeQuery()->fetchAll(\PDO::FETCH_COLUMN);
}
$paths = array_merge(...$chunks);
```

come-nc: According to what I could find and test, you’re right, array_push with the spread is better.

PR #4240, opened 2026-01-05
chore(dev-deps): Update psalm to support PHP 8.5

come-nc:


come-nc:
I cannot get it to be picked up by a custom runner to test with 8.5 :cry:

come-nc:
Failing cypress is not related, I think it’s related to https://github.com/nextcloud/server/pull/57277

susnux:
@come-nc cypress fix in https://github.com/nextcloud/groupfolders/pull/4241

PR #4262, opened 2026-01-14
feat(MountProvider): Implement IPartialMountProvider

provokateurin:
I was not able to really test it, because the codepath is never triggered (without any modifications in SetupManager), but from step debugging the output looked good to me.

From https://github.com/nextcloud/server/pull/57497 I gathered that I don't even need to iterate the `$mountProviderArgs`, so maybe that loop could be removed.

Also this method doesn't implement the conflict resolution, which might be necessary to add.

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -8,6 +8,8 @@ declare (strict_types=1);
 
 namespace OCA\GroupFolders\Mount;
 
+use Exception;
+use OC;
 use OC\Files\Storage\Wrapper\PermissionsMask;
 use OCA\GroupFolders\ACL\ACLManager;
 use OCA\GroupFolders\ACL\ACLManagerFactory;
@@ -19,6 +21,8 @@ use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Config\IMountProvider;
 use OCP\Files\Config\IMountProviderCollection;
+use OCP\Files\Config\IPartialMountProvider;
+use OCP\Files\Config\MountProviderArgs;
 use OCP\Files\Mount\IMountPoint;
 use OCP\Files\Storage\IStorage;
 use OCP\Files\Storage\IStorageFactory;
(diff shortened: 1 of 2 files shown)


lib/Mount/MountProvider.php
```diff
@@ -264,4 +259,46 @@ private function findConflictsForUser(IUser $user, array $mountPoints): array {
...
+			$userFolders[$user->getUID()] ??= $this->folderManager->getFoldersForUser($user);
+			$folders = $userFolders[$user->getUID()];
+
+			foreach ($folders as $folder) {
+				$mountPoint = '/' . $user->getUID() . '/files/' . $folder->mountPoint;
+				if (isset($mounts[$mountPoint])) {
+					continue;
+				}
+
+				if ((!$forChildren && $mountProviderArg->mountInfo->getMountPoint() === $folder->mountPoint) || ($forChildren && str_starts_with($folder->mountPoint, $mountProviderArg->mountInfo->getMountPoint() . '/'))) {
```
icewind1991: `$folder->mountPoint` is relative to the users home, `$mountProviderArg->mountInfo->getMountPoint()` is absolute

lib/Mount/MountProvider.php
```diff
@@ -264,4 +259,46 @@ private function findConflictsForUser(IUser $user, array $mountPoints): array {
...
+		/** @var array<string, list<FolderDefinitionWithPermissions>> $userFolders */
+		$userFolders = [];
+		/** @var array<string, ACLManager> $userAclManagers */
+		$userAclManagers = [];
+
+		/** @var IMountProviderArgs $mountProviderArg */
+		foreach ($mountProviderArgs as $mountProviderArg) {
+			$user = $mountProviderArg->mountInfo->getUser();
+
+			$userFolders[$user->getUID()] ??= $this->folderManager->getFoldersForUser($user);
```
icewind1991: Without passing the filtering down to this call this doesn't really do much for performance.

Ideally it would also pass down the `CacheEntry`s for the mounts from the `MountProviderArgs` to prevent having to re-fetch those. But that can be done in a follow-up if desired.

lib/Mount/MountProvider.php
```diff
@@ -61,7 +61,7 @@ public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
 		$aclManager = $this->aclManagerFactory->getACLManager($user);
 		$rootRules = $aclManager->getRulesByFileIds($rootFileIds);
 
-		return array_filter(array_map(function (FolderDefinitionWithPermissions $folder) use ($user, $loader, $conflicts, $aclManager, $rootRules): ?IMountPoint {
+		return array_filter(array_map(function (FolderDefinitionWithPermissions $folder) use ($user, $loader, $conflicts, $aclManager, $rootRules): IMountPoint {
```
come-nc: If the return type cannot be null anymore I suppose the array_filter is not needed anymore either?

PR #4263, opened 2026-01-14
perf: Move folder pagination to the database

provokateurin:
~~Ordering is still done in PHP, as we need to make sure that the order is always the same (and not DB specific).~~

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -136,10 +136,24 @@ class FolderManager {
 	 * @return array<int, FolderWithMappingsAndCache>
 	 * @throws Exception
 	 */
-	public function getAllFoldersWithSize(): array {
+	public function getAllFoldersWithSize(int $offset = 0, ?int $limit = null, string $orderBy = 'mount_point', string $order = 'ASC'): array {
 		$applicableMap = $this->getAllApplicable();
 
 		$query = $this->selectWithFileCache();
+		$query->setFirstResult($offset);
+		$query->setMaxResults($limit);
+		if ($orderBy === 'groups') {
+			$query
+				->leftJoin('f', 'group_folders_groups', 'g', $query->expr()->eq('f.folder_id', 'g.folder_id'))
+				->groupBy('f.folder_id')
+				->orderBy($query->func()->count('g.applicable_id'), $order);
+		} else {
+			$query->orderBy($orderBy, $order);
+		}
+		// Fallback in case two rows are the same after ordering by the $orderBy
+		if ($orderBy !== 'mount_point') {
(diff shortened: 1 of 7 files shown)


icewind1991:
I don't think paginating in the DB but sorting in PHP will work. If the enter in the DB are `A`, `C`, `B` in that order, we paginate trough the folders with a page-size of 2, the results will not be ordered. (and still DB specific)

provokateurin:
Oops yes, I'm too tired today :sweat_smile:

provokateurin:
~~I moved ordering to the DB and removed all ordering in the frontend, because we need to have a single source of truth. It works correctly, except for changing the order when seeing a different page than the first one. I'll fix that as well, but should be quite minor. The rest can already be reviewed as is.~~

Today I can't reproduce this issue anymore.

lib/Folder/FolderManager.php
```diff
@@ -136,10 +136,31 @@ private function selectWithFileCache(?IQueryBuilder $query = null): IQueryBuilde
 	 * @return array<int, FolderWithMappingsAndCache>
 	 * @throws Exception
 	 */
-	public function getAllFoldersWithSize(): array {
+	public function getAllFoldersWithSize(?int $offset = null, ?int $limit = null, ?string $orderBy = null, ?string $order = null): array {
```
icewind1991: `string $orderBy = 'mount_point', string $order = 'ASC'`

lib/Folder/FolderManager.php
```diff
@@ -136,10 +136,31 @@ private function selectWithFileCache(?IQueryBuilder $query = null): IQueryBuilde
...
+			$query
+				->leftJoin('f', 'group_folders_groups', 'g', $query->expr()->eq('f.folder_id', 'g.folder_id'))
+				->groupBy('f.folder_id')
+				->orderBy($query->func()->count('g.applicable_id'), $order);
+		} else {
+			$query->orderBy($orderBy, $order);
+		}
+
+		if ($orderBy !== 'mount_point') {
+			$query->addOrderBy('mount_point', 'ASC');
```
icewind1991, line 155: Don't we still need this to ensure consistent results between databases?

provokateurin: If there is a difference between DBs then it would be nice, but how would you do it?

provokateurin: Or maybe I didn't get your question? This statement is used as a fallback, for the case that two rows are in the same order if they they are ordered by a different column, so they are sorted by the mount_point as well.

icewind1991: Right, I misread the code, ignore my question

PR #4283, opened 2026-01-22
fix: Optimize group folder queries

Rello:
### Motivation
- Avoid full-table scans and high-memory iterations when checking mountpoints by using a targeted lookup instead of iterating all folders.  
- Reduce global table reads for applicable/group mappings by scoping queries to only requested folder IDs.  

### Description
- Replace loop-based mountpoint existence check in `lib/Controller/FolderController.php` with a call to `FolderManager::mountPointExists()` to perform a single targeted query.  
- Add `mountPointExists()` to `lib/Folder/FolderManager.php` to check for a mount point using a single-row query.  
- Introduce `getApplicableForFolders()` and `getApplicableForFolder()` and `getFolderMappingsForFolders()` in `lib/Folder/FolderManager.php` to fetch applicable groups/circles and manage mappings for a specified list of folder IDs.  
- Update `getAllFoldersWithSize()`, `getAllFoldersForUserWithSize()`, `getFolder()`, `getGroups()`, and `getCircles()` to use the new scoped helpers instead of loading all applicable/mapping rows, including safe early-return when the folder ID list is empty.  

### Background
- customer performance issue with >50k group folders
- required for NC31


lib/Folder/FolderManager.php
```diff
@@ -241,6 +242,36 @@ private function getAllFolderMappings(): array {
...
+
+		$folderMap = [];
+		foreach ($rows as $row) {
+			$id = (int)$row['folder_id'];
+
+			if (!isset($folderMap[$id])) {
+				$folderMap[$id] = [$row];
+			} else {
+				$folderMap[$id][] = $row;
+			}
```
```suggestion
			$folderMap[$id] ??= [];
			$folderMap[$id][] = $row;
```

lib/Folder/FolderManager.php
```diff
@@ -404,12 +434,76 @@ private function getAllApplicable(): array {
...
+
+		$queryHelper?->addCircleDetails('g', 'circle_id');
+
+		$rows = $query->executeQuery()->fetchAll();
+		$applicableMap = [];
+		foreach ($rows as $row) {
+			$id = (int)$row['folder_id'];
+			if (!array_key_exists($id, $applicableMap)) {
+				$applicableMap[$id] = [];
+			}
```
```suggestion
			$applicableMap[$id] ??= [];
```

lib/Folder/FolderManager.php
```diff
@@ -404,12 +434,76 @@ private function getAllApplicable(): array {
...
+
+			if (!$row['circle_id']) {
```
provokateurin: This check is very ugly, at least invert it

lib/Folder/FolderManager.php
```diff
@@ -474,6 +568,20 @@ public function canManageACL(int $folderId, IUser $user): bool {
...
+	public function mountPointExists(string $mountPoint): bool {
+		$query = $this->connection->getQueryBuilder();
+		$query->select('folder_id')
+			->from('group_folders')
+			->where($query->expr()->eq('mount_point', $query->createNamedParameter($mountPoint)))
+			->setMaxResults(1);
+
+		$result = $query->executeQuery();
+		$exists = $result->fetchOne() !== false;
+		$result->closeCursor();
```
provokateurin: Selecting the count of rows might allow the DB to operate more efficiently.

Rello: Updated the mount point existence check to use a count-based query for a more direct DB existence check

lib/Folder/FolderManager.php
Rello: Hi,
this is the same in the "original" function to get all folders. I would guess the idea should remain the same for consistency

https://github.com/nextcloud/groupfolders/blob/master/lib/Folder/FolderManager.php#L378

provokateurin: Then just adjust both methods. It's better to fix it than to duplicate it.

Rello: I changed it. let me know..

lib/Folder/FolderManager.php
```diff
@@ -241,6 +242,33 @@ private function getAllFolderMappings(): array {
 		return $folderMap;
 	}
 
+	/**
+	 * @return array<int, list<InternalFolderMapping>>
+	 * @throws Exception
+	 */
+	private function getFolderMappingsForFolders(array $folderIds): array {
```
provokateurin, line 255: Instead of duplicating most of the code of this method, please add it to the existing method with an optional parameter like you did it for the other method.

lib/Folder/FolderManager.php
```diff
@@ -375,15 +409,7 @@ private function getAllApplicable(): array {
...
-			if (!$row['circle_id']) {
-				$entityId = (string)$row['group_id'];
-
-				$entry = [
-					'displayName' => $this->groupManager->get($row['group_id'])?->getDisplayName() ?? $row['group_id'],
-					'permissions' => (int)$row['permissions'],
-					'type' => 'group',
-				];
-			} else {
+			if ($row['circle_id']) {
```
```suggestion
			if ($row['circle_id'] !== null) {
```

lib/Folder/FolderManager.php
```diff
@@ -241,6 +248,14 @@ private function getAllFolderMappings(): array {
 		return $folderMap;
 	}
 
+	/**
+	 * @return array<int, list<InternalFolderMapping>>
+	 * @throws Exception
+	 */
+	private function getFolderMappingsForFolders(array $folderIds): array {
+		return $this->getAllFolderMappings($folderIds);
+	}
```
provokateurin, line 257: This method is useless, please remove it.

lib/Folder/FolderManager.php
```diff
@@ -401,6 +412,14 @@ private function getAllApplicable(): array {
 					'permissions' => (int)$row['permissions'],
 					'type' => 'circle',
 				];
+			} else {
+				$entityId = (string)$row['group_id'];
+
+				$entry = [
+					'displayName' => $this->groupManager->get($row['group_id'])?->getDisplayName() ?? $row['group_id'],
```
provokateurin, line 419: Your rebase went wrong, you removed group displayname caching which was merged just today. Please see the negative diff above.

lib/Folder/FolderManager.php
```diff
@@ -157,7 +155,9 @@ public function getAllFoldersWithSize(int $offset = 0, ?int $limit = null, strin
 
 		$rows = $query->executeQuery()->fetchAll();
 
-		$folderMappings = $this->getAllFolderMappings();
+		$folderIds = array_values(array_map(static fn (array $row): int => (int)$row['folder_id'], $rows));
+		$applicableMap = $this->getApplicableForFolders($folderIds);
+		$folderMappings = $this->getFolderMappingsForFolders($folderIds);
```
provokateurin, line 160: TBH I wonder if it would even be possible to do this with a LEFT JOIN and only have a single query.

lib/Folder/FolderManager.php
```diff
@@ -409,12 +428,29 @@ private function getAllApplicable(): array {
...
+	}
+
+	/**
+	 * @param list<int> $folderIds
+	 * @return array<int, array<string, GroupFoldersApplicable>>
+	 * @throws Exception
+	 */
+	private function getApplicableForFolders(array $folderIds): array {
+		return $this->getAllApplicable($folderIds);
+	}
```
provokateurin, line 446: Also useless

Rello:
I will close it and do a clean one due to the frequent back & forth

PR #4289, opened 2026-01-23
fix(files): @nextcloud/files usage

artonge:
`@nc/vue` is now a pure ES module, so groupfolder bundling need to be adapted :see_no_evil:, don't have enough time to fight with npm and webpack


susnux:
* resolved with #4301

PR #4316, opened 2026-01-28
perf(FolderManager): Cache group display name

CarlSchwan:
Otherwise it can be quickly called way too many times which generate requests to the group backends.

image

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -381,12 +381,7 @@ class FolderManager {
 			if (!$row['circle_id']) {
 				$entityId = (string)$row['group_id'];
 
-				if (isset($groupDisplayNameCache[$row['group_id']])) {
-					$displayName = $groupDisplayNameCache[$row['group_id']];
-				} else {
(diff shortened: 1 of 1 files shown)


lib/Folder/FolderManager.php
```diff
@@ -378,8 +381,15 @@ private function getAllApplicable(): array {
 			if (!$row['circle_id']) {
 				$entityId = (string)$row['group_id'];
 
+				if (isset($groupDisplayNameCache[$row['group_id']])) {
+					$displayName = $groupDisplayNameCache[$row['group_id']];
+				} else {
+					$displayName = $this->groupManager->get($row['group_id'])?->getDisplayName() ?? $row['group_id']
+					$groupDisplayNameCache[$row['group_id']] = $displayName;
+				}
```
```suggestion
				$displayName = $groupDisplayNameCache[$entityId] ??= this->groupManager->get($entityId)?->getDisplayName() ?? $entityId;
```
https://3v4l.org/4jnvK#vnull

provokateurin (changes requested):

Issue #4095, opened 2025-10-16
Performance regression 19 vs. 20


SystemKeeper:
* Follow-up to https://github.com/nextcloud/groupfolders/issues/4073

Doing the same test as mentioned in the issue above but with https://github.com/nextcloud/groupfolders/pull/4075 applied sees a halving in the query count (as expected).

Then I downgraded versions and compared 19.1.5 to 21.1.1.:
Query count on 21.1.1 (with patch): ~600 queries (so basically 1 per groupfolder)
Query count on 19.1.5: ~50 queries.

Bisect shows:

```
2637d7db95e854a942579d3f2914894e8f05951b is the first bad commit
commit 2637d7db95e854a942579d3f2914894e8f05951b
Author: Robin Appelman <robin@icewind.nl>
Date:   Thu Jun 19 16:59:09 2025 +0200

    fix: remove assumptions that all groupfolders share a storage id
    
    Signed-off-by: Robin Appelman <robin@icewind.nl>
```

provokateurin:
Also another thing, not sure if related to the new version: https://github.com/nextcloud/groupfolders/issues/4143

tpelingan:
Hi @provokateurin we created a suggested fix here - https://github.com/nextcloud/groupfolders/pull/4676

PR #4320, opened 2026-01-29
Performance optimisation for high number of team folders

Rello:
### Motivation
- Avoid full-table scans and high-memory iterations when checking mountpoints by using a targeted lookup instead of iterating all folders.  
- Reduce global table reads for applicable/group mappings by scoping queries to only requested folder IDs.  

### Description
- Replace loop-based mountpoint existence check in `lib/Controller/FolderController.php` with a call to `FolderManager::mountPointExists()` to perform a single targeted query.  
- Add `mountPointExists()` to `lib/Folder/FolderManager.php` to check for a mount point using a single-row query.  
- Introduce `getApplicableForFolders()` and `getApplicableForFolder()` and `getFolderMappingsForFolders()` in `lib/Folder/FolderManager.php` to fetch applicable groups/circles and manage mappings for a specified list of folder IDs.  
- Update `getAllFoldersWithSize()`, `getAllFoldersForUserWithSize()`, `getFolder()`, `getGroups()`, and `getCircles()` to use the new scoped helpers instead of loading all applicable/mapping rows, including safe early-return when the folder ID list is empty.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -158,7 +158,7 @@ class FolderManager {
 		$folderIds = array_values(array_map(static fn (array $row): int => (int)$row['folder_id'], $rows));
 		$applicableMap = $this->getAllApplicable($folderIds);
 		$folderMappings = $this->getAllFolderMappings($folderIds);
-		
+
 		$folderMap = [];
 		foreach ($rows as $row) {
 			$folder = $this->rowToFolder($row);


SystemKeeper:
Would that fix the regression in https://github.com/nextcloud/groupfolders/issues/4095 ?

provokateurin:
It could partly, but I think all changes here were already applicable to earlier versions before the "bad" commit.

Rello:
Hello,

I have another update call with the customer today evening. They will need this backported to 31.
What timeline is realistic here?

PR #4344, opened 2026-02-05
PHPStan

provokateurin:
TODO:
- [x] https://github.com/nextcloud/server/pull/58098
- [x] https://github.com/nextcloud/circles/pull/2337
- [x] https://github.com/nextcloud/server/pull/58261
- [x] https://github.com/nextcloud/server/pull/58256
- [x] https://github.com/nextcloud/server/pull/58355
- [x] https://github.com/nextcloud/server/pull/58286
- [x] https://github.com/icewind1991/nextcloud-version-matrix/pull/3
- [ ] Submit and backport bugs that were found separately
- [x] Create GHA workflow
- [x] Increase level even further


lib/Migration/Version104000Date20180918132853.php
```diff
@@ -23,6 +23,9 @@ public function description(): string {
 		return 'Adds table to store trashbin information for Team folders';
 	}
 
+	/**
+	 * @phpstan-ignore missingType.iterableValue
+	 */
```
CarlSchwan: https://github.com/nextcloud/server/pull/58282 should help

provokateurin: Very easy to convince me to review your PR this way 😁

provokateurin:
Still some things to do, but the code changes can be reviewed already.

CarlSchwan:
I'm not really a fan of removing the stubs. We should make it harder to use private apis not simpler.

provokateurin:
> I'm not really a fan of removing the stubs. We should make it harder to use private apis not simpler.

I get your point, but I'm not sure if this is the right way.

Anyway PHPStan has support for stubs, but doesn't like the currently generated ones. I tried a bit, but wasn't able to generate something it accepted. I will have another look again, as this also makes CI easier (no need to clone server and circles).

PR #4347, opened 2026-02-06
fix: set displayname for teamfolders

benjaminfrueh:
### Problem
Teamfolders were displayed using their folder ids instead of their names in the teamfolders view.

### Solution
Move `displayname` from `attributes` to be a top-level property in `nodeData`


provokateurin:
stable31 is not affected

PR #4348, opened 2026-02-06
feat: validate mountpoint using FilenameValidator

benjaminfrueh:
### Problem
Teamfolder with invalid characters (particularly `/` and `\`) in their mount point cause issues and errors.

### Solution
Implement validation using `IFilenameValidator` to reject invalid mount point names at creation and rename and display the error.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -28,7 +28,9 @@ use OCP\DB\Exception;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\FileInfo;
+use OCP\Files\IFilenameValidator;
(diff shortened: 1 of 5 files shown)


lib/Folder/FolderManager.php
```diff
@@ -729,11 +732,19 @@ public function getFoldersFromCircleMemberships(IUser $user, ?int $folderId = nu
 		}, $query->executeQuery()->fetchAll()));
 	}
 
+	/**
+	 * @throws InvalidPathException
+	 */
+	private function validateMountPoint(string $mountPoint): void {
+		$this->filenameValidator->validateFilename($mountPoint);
```
provokateurin, line 739: We do allow `/` though, so that needs to be kept.

provokateurin:
Replaced by https://github.com/nextcloud/groupfolders/pull/4368

PR #4360, opened 2026-02-09
fix(trash): Hide trash from admins

provokateurin:
Instance admins can currently see the trash of all groupfolders, making their trashbin unusable. This is NOT a security issue because an admin could just add themselves to a groupfolder in order to see the trash.


julien-nc:
I can confirm this works. With this patch, if there an ACL denying all access rights to an admin (or the admin group) for a subfolder, the admins don't see the deleted files in their trashbin.
Without this patch, they do.

PR #4368, opened 2026-02-10
fix(FolderController): Sanitize mountpoints correctly

provokateurin:
Team folders with leading or trailing slashes were never shown in the user's files. Using a backslash in the path was fine, but it was internally converted to a slash anyway, so now it's also saved as a slash.

### DIFF
--- a/src/settings/App.tsx
+++ b/src/settings/App.tsx
@@ -18,6 +18,14 @@ import AdminGroupSelect from './AdminGroupSelect'
 import SubAdminGroupSelect from './SubAdminGroupSelect'
 import { loadState } from '@nextcloud/initial-state'
 import { t } from '@nextcloud/l10n'
+import { isAxiosError } from "@nextcloud/axios";
+import { showError } from "@nextcloud/dialogs";
+import { getLoggerBuilder } from '@nextcloud/logger'
+
+const logger = getLoggerBuilder()
+	.setApp('groupfolders')
+	.detectUser()
+	.build()
 
 const bytesInOneGibibyte = Math.pow(1024, 3)
 const defaultQuotaOptions = {
@@ -98,10 +106,20 @@ export class App extends Component<unknown, AppState> implements OC.Plugin<OC.Se
 		if (!mountPoint) {
 			return
 		}
-		const folder = await this.api.createFolder(mountPoint, this.state.newACLDefaultNoPermission)
-		const folders = this.state.folders
-		folders.push(folder)
-		this.setState({ folders, newMountPoint: '' })
+		try {
+			const folder = await this.api.createFolder(mountPoint, this.state.newACLDefaultNoPermission);
+			const folders = this.state.folders
+			folders.push(folder)
+			this.setState({folders, newMountPoint: ''})
+		} catch (error) {
+			logger.error('Error while creating new folder', { error })
+
+			if (isAxiosError(error) && error.response.data.message) {
+				showError(error.response.data.message)
+			} else {
+				showError(t('groupfolders', 'Folder could not be created'))
+			}
+		}
 	}
 
 	attach = (search: OC.Search.Core) => {
@@ -161,10 +179,20 @@ export class App extends Component<unknown, AppState> implements OC.Plugin<OC.Se
 	}
 
 	async renameFolder(folder: Folder, newName: string) {
-		await this.api.renameFolder(folder.id, newName)
-		const folders = this.state.folders
-		folder.mount_point = newName
-		this.setState({ folders, editingMountPoint: 0 })
+		try {
+			await this.api.renameFolder(folder.id, newName);
+			const folders = this.state.folders
+			folder.mount_point = newName
+			this.setState({folders, editingMountPoint: 0})
+		} catch (error) {
+			logger.error('Error while renaming folder', { error })
+
+			if (isAxiosError(error) && error.response.data.message) {
+				showError(error.response.data.message)
+			} else {
+				showError(t('groupfolders', 'Folder could not be renamed'))
+			}
+		}
 	}
 
 	async setAcl(folder: Folder, acl: boolean) {


benjaminfrueh (approved):

lib/Controller/FolderController.php
```diff
@@ -235,24 +236,44 @@ private function getRootFolderStorageId(): ?int {
...
 		$folder = $this->checkedGetFolder($id);
 
 		return new DataResponse($this->formatFolder($folder));
 	}
 
+	private function trimMountpoint(string $mountpoint): string {
+		$mountpoint = FileSystem::normalizePath($mountpoint);
+
+		// Remove leading slash
+		$mountpoint = substr($mountpoint, 1);
```
artonge: We usually do the following:


```suggestion
		$mountpoint = rtrim($mountpoint, '/')
```

lib/Controller/FolderController.php
```diff
@@ -235,24 +236,44 @@ private function getRootFolderStorageId(): ?int {
...
+
+		if (str_contains($mountpoint, '..')) {
+			throw new OCSBadRequestException('Mount point cannot contain ".."');
+		}
+
+		// Check if the new mountpoint is valid
+		if ($mountpoint === '') {
```
artonge: If the mountpoint originally is `'/'`, then after the trim it will be `''`, so this is valid.

artonge: Maybe add

```php
if ($mountpoint === '') {
    $mountpoint = '/';
}
```

lib/Controller/FolderController.php
artonge: Also, would be good to keep
```php
trim($mountpoint);
```
So that whitespace are still removed.

src/settings/App.tsx
```diff
@@ -98,10 +106,19 @@ export class App extends Component<unknown, AppState> implements OC.Plugin<OC.Se
...
+			this.setState({folders, newMountPoint: ''})
+		}).catch((error) => {
+			logger.error('Error while creating new folder', { error })
+
+			if (isAxiosError(error) && error.response.data.message) {
+				showError(error.response.data.message)
+			} else {
+				showError(t('groupfolders', 'Folder could not be created'))
+			}
+		})
```
artonge: Maybe use the `await` syntax:

```suggestion
		try {
			const folder = await this.api.createFolder(mountPoint, this.state.newACLDefaultNoPermission)
			const folders = this.state.folders
			folders.push(folder)
			this.setState({folders, newMountPoint: ''})
		} catch (error) {
			logger.error('Error while creating new folder', { error })

			if (isAxiosError(error) && error.response.data.message) {
				showError(error.response.data.message)
			} else {
				showError(t('groupfolders', 'Folder could not be created'))
			}
		}
```

src/settings/App.tsx
```diff
@@ -161,10 +178,19 @@ export class App extends Component<unknown, AppState> implements OC.Plugin<OC.Se
...
+			this.setState({folders, editingMountPoint: 0})
+		}).catch((error) => {
+			logger.error('Error while renaming folder', { error })
+
+			if (isAxiosError(error) && error.response.data.message) {
+				showError(error.response.data.message)
+			} else {
+				showError(t('groupfolders', 'Folder could not be renamed'))
+			}
+		})
```
artonge: Same as above

lib/Controller/FolderController.php
icewind1991: Though [currently broken](https://github.com/nextcloud/groupfolders/pull/4382), we do support having groupfolders mounted at the users home directory.

So we'll need to set it back to `/` in this case

provokateurin: I wouldn't reset it, because the mountpoint could also be empty for other reasons, but I get what you mean.

lib/Controller/FolderController.php
provokateurin: `ltrim` would be the correct method, but TBH I don't see any benefit. In this place we know there is always exactly on leading slash, so it's faster to just remove that than to call `ltrim` which needs to compare characters first.
I'm adding back the `trim` though.

lib/Controller/FolderController.php
provokateurin: Needs to be done earlier, because it can be empty for other reasons as well.

src/settings/App.tsx
provokateurin: How is this any better?

provokateurin:
Verified the use case with `/` as the mountpoint still works and also made the commands sanitize the mountpoint correctly.

lib/Controller/FolderController.php
artonge: > I don't see any benefit

I would do it for safety and consistency. But I agree that it is not necessary in this place.

src/settings/App.tsx
artonge: Easier to read, and matches what we are doing in other places.

artonge (approved):

src/settings/App.tsx
provokateurin: > matches what we are doing in other places

Not in this app, we almost exclusively use then+catch instead of try+catch.

provokateurin: And it's really not easier to read, check this diff: https://github.com/nextcloud/groupfolders/compare/140d251df98de38db68ff2db301447bd19a99c1a..4d4ecb4ae28f7564bba81979a6fd0b47e0d7826a

Anyway I changed it, so you can approve the PR :joy:

lib/Folder/FolderManager.php
```diff
@@ -721,6 +723,31 @@ public function getFoldersFromCircleMemberships(IUser $user, ?int $folderId = nu
...
+
+		// Remove leading slash
+		$mountpoint = substr($mountpoint, 1);
+
+		if (str_contains($mountpoint, '..')) {
+			throw new OCSBadRequestException('Mount point cannot contain ".."');
+		}
+
+		// Check if the new mountpoint is valid
+		if ($mountpoint === '') {
```
artonge, line 745: If you call ` FileSystem::normalizePath($mountpoint);` I don't think this will ever be `true`

provokateurin: It happens when you use a `.` as the mountpoint. There are might be other cases where it could happen as well.

Issue #4377, opened 2026-02-10
OCC :list - Group and Quotas columns are inverted


La-Fabrique-Info:
Hi,
As the title says, Group and Quotas columns are inverted when using groupfolder occ command.

### Steps to reproduce
1. execute: `php occ groupfolders:list`

### Expected behaviour
+---------------+--------------------+----------------+--------------------------------------------------+
| Folder Id | Name              | Quota       | Groups                                                |...
+--------------+---------------------+----------------+--------------------------------------------------+
| 1              | Folder_Name | Unlimited | GROUP: read, write, share, delete | ...
+--------------+---------------------+----------------+--------------------------------------------------+

### Actual behaviour
+---------------+--------------------+------------------+-------------------------------------------------+
| Folder Id | Name              | Groups       | Quotas                                               |...
+--------------+---------------------+-----------------+--------------------------------------------------+
| 1              | Folder_Name | Unlimited   | GROUP: read, write, share, delete | ...
+--------------+---------------------+-----------------+--------------------------------------------------+

### Server configuration
**Nextcloud version:** 32.0.5
**Team Folder version:** 20.1.10

provokateurin:
Good catch!

provokateurin:
https://github.com/nextcloud/groupfolders/pull/4378

PR #4378, opened 2026-02-10
fix(ListCommand): Align quota and groups column order

provokateurin:
Fixes https://github.com/nextcloud/groupfolders/issues/4377


provokateurin:
stable31 is not affected.

PR #4396, opened 2026-02-11
feat: performance tuning for ACL rule derivation

Rello:
Optimized ACLManager::getRelevantRulesForPath() by replacing repeated array_merge() + array_unique() calls with map-style deduplication ($allPaths[$path] = true) and passing array_keys() to getRules(). This avoids repeated large-array rebuilds

_Blackfire Profiling on NC31 with 30k Team Folders + 30k ACLs_

Before: 
Bildschirmfoto 2026-02-11 um 15 06 19

After:
Bildschirmfoto 2026-02-11 um 15 05 49

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -164,7 +164,7 @@ class ACLManager {
 			}
 		}
 
-		return $this->calculatePermissionsForPath($filteredRules);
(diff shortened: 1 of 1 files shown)


Rello:
`getPermissionsForPathFromRules` was reduced from 2.5s to 0.3s

PR #4404, opened 2026-02-12
Add deletion of group_folders_groups in FolderManager

kevin-256:
Automatically removes associated permissions upon groupfolder deletion to maintain database consistency and cleanliness.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -939,7 +939,7 @@ class FolderManager {
 					),
 				);
 			$query->executeStatement();
-			
+
 			$query = $this->connection->getQueryBuilder();
 			$query->delete('group_folders_groups')
 				->where(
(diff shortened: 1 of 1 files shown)


provokateurin (commented):
Looks good, but if you don't mind, could you check the other tables as well? They don't seem to be cleaned up either, so this would be nice for completeness.

kevin-256:
I've checked the other tables and deleted entries related to that folder. The only entries I haven't deleted are those in the `oc_group_folders_versions` table, which remain when deleting a folder containing a file.
Steps to reproduce:
- Create a folder.
- Create an MD file inside it.
- Edit the file.
- Delete the folder.

The same issue occurs with 'normal' folders and 'normal' files outside group folders, so I think the problem is in the Files app. These entries may be kept in the database for logging purposes, so I'm not sure if deleting data from oc_group_folders_trash is appropriate.

lib/Folder/FolderManager.php
```diff
@@ -889,15 +882,40 @@ public function setManageACL(int $folderId, string $type, string $id, bool $mana
...
 
-		$this->updateOverwriteHomeFolders();
+			$this->connection->commit();
+			
+			$this->eventDispatcher->dispatchTyped(new CriticalActionPerformedEvent('The groupfolder with id %d was removed', [$folderId]));
+	
+			$this->updateOverwriteHomeFolders();
+		} catch (Exception $e) {
+            $this->connection->rollBack();
+			$this->logger->notice('Error while removing groupfolder', ['exception' => $e]);
```
```suggestion
			$this->logger->error('Error while removing groupfolder', ['exception' => $e]);
```

provokateurin (approved):
Thanks!

PR #4429, opened 2026-02-19
fix(ACL): apply ACL rules in the correct order

CarlSchwan:
Before this change the filteredRules array would be flipped in comparison to the $rules. So rules for the parents would be applied before rules for the children.

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -164,6 +164,8 @@ class ACLManager {
 			}
 		}
 
+		uksort($filteredRules, static fn (string $a, string $b) => strlen($a) <=> strlen($b));
+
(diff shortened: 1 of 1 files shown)


lib/ACL/ACLManager.php
```diff
@@ -164,6 +164,8 @@ public function getPermissionsForPathFromRules(int $folderId, string $path, arra
 			}
 		}
 
+		uksort($filteredRules, static fn(string $a, string $b) => \strlen($a) <=> \strlen($b));
```
provokateurin:
```suggestion
		uksort($filteredRules, static fn(string $a, string $b) => strlen($a) <=> strlen($b));
```

salmart-dev (approved):

PR #4451, opened 2026-02-23
perf(MountProvider): optimize and improve conflict handling

solracsf:
Function was called inside `array_map` for every conflicting folder; this dropped from N calls to 1.
Reduce mount resolution time from ~50-150ms to ~5-20ms per call.

### DIFF
--- a/lib/Mount/MountProvider.php
+++ b/lib/Mount/MountProvider.php
@@ -54,43 +54,64 @@ class MountProvider implements IMountProvider, IPartialMountProvider {
 	public function getMountsForUser(IUser $user, IStorageFactory $loader): array {
 		$folders = $this->folderManager->getFoldersForUser($user);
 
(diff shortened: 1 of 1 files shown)


provokateurin:
Are you referring to `$userStorage = $this->mountProviderCollection->getHomeMountForUser($user)->getStorage();`? I fail to see any other significant changes.

solracsf:
Fewer passes over `$folders`. The old code did `array_map` (mount points) + `array_filter` (ACL folders) + `array_map` (root ids) + `array_map` (build mounts), 4 iterations. New code does 2. Clean win.

Lazy `$userStorage`. Previously `getHomeMountForUser()` was invoked once per conflicting folder inside the `array_map` callback. Now once total, and only if a conflict exists. Real improvement for users with several conflicts.

Issue #4442, opened 2026-02-21
IPartialMountProvider breaks nested subfolders


SystemKeeper:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
Create a folder structure like this:

<img width="1114" height="300" alt="Image" src="https://github.com/user-attachments/assets/1a68c29a-34ef-432e-88c1-e9c09b4da977" />

### Expected behaviour
When navigating to the top level folder, it should show 2 sub level folders

### Actual behaviour

<img width="512" height="235" alt="Image" src="https://github.com/user-attachments/assets/20251be3-9232-483a-8beb-814f937a3a0b" />

This happens on stable33 and master. Running bisect on master shows:

```
706db843295facc67d425093951058556317b622 is the first bad commit
commit 706db843295facc67d425093951058556317b622 (HEAD)
Author: provokateurin <kate@provokateurin.de>
Date:   Wed Jan 14 13:35:49 2026 +0100

    feat(MountProvider): Implement IPartialMountProvider
    
    Signed-off-by: provokateurin <kate@provokateurin.de>

 lib/Folder/FolderManager.php | 26 +++++++++++++++++++++-----
 lib/Mount/MountProvider.php  | 81 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++------------------
 2 files changed, 84 insertions(+), 23 deletions(-)
```

Downgrading to v21.0.2 also solves the issue on 33 (last version without partial mount provider.

@provokateurin

SystemKeeper:
https://github.com/nextcloud/groupfolders/blob/3a1493f50d680b501d2c6bb1da2ddcb9a692e09c/lib/Mount/MountProvider.php#L291

`$mountProviderArgs` contains both mount points:

<img width="1292" height="429" alt="Image" src="https://github.com/user-attachments/assets/08c934e6-a85f-49a8-a9c5-8143fc577d62" />

So the above mentioned line will only work for the first folder.

SystemKeeper:
[...]

PR #4453, opened 2026-02-23
feat: Combine getFoldersForGroups database calls when paths are known

SystemKeeper:
* Ref https://github.com/nextcloud/groupfolders/issues/4442 and https://github.com/nextcloud/groupfolders/pull/4448#pullrequestreview-3840271076

Instead of querying each folder that we receive through `$mountProviderArgs` separately, we try to do it in a single query now.

I think it should be basically this, but please take a close look.

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -636,7 +636,7 @@ class FolderManager {
 	 * @return list<FolderDefinitionWithPermissions>
 	 * @throws Exception
 	 */
-	public function getFoldersForGroups(array $groupIds, ?int $folderId = null, ?string $path = null): array {
+	public function getFoldersForGroups(array $groupIds, ?int $folderId = null, ?array $paths = null): array {
 		if (count($groupIds) === 0) {
 			return [];
 		}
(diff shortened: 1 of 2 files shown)


lib/Folder/FolderManager.php
```diff
@@ -708,8 +719,8 @@ public function getFoldersFromCircleMemberships(IUser $user, ?int $folderId = nu
 			$query->andWhere($query->expr()->eq('f.folder_id', $query->createNamedParameter($folderId, IQueryBuilder::PARAM_INT)));
 		}
 
-		if ($path !== null) {
-			$query->andWhere($query->expr()->eq('f.mount_point', $query->createNamedParameter($path)));
+		if ($paths !== null) {
+			$query->andWhere($query->expr()->in('f.mount_point', $query->createNamedParameter($paths, IQueryBuilder::PARAM_INT_ARRAY)));
```
provokateurin: This also needs chunking, no?

```suggestion
			$query->andWhere($query->expr()->in('f.mount_point', $query->createNamedParameter($paths, IQueryBuilder::PARAM_STR_ARRAY)));
```

lib/Mount/MountProvider.php
```diff
@@ -267,16 +267,14 @@ private function findConflictsForUser(IUser $user, array $mountPoints): array {
...
 	#[Override]
 	public function getMountsForPath(string $setupPathHint, bool $forChildren, array $mountProviderArgs, IStorageFactory $loader): array {
-		/** @var array<string, IMountPoint> $mounts */
-		$mounts = [];
+		/** @var IUser $user */
+		$user = $mountProviderArgs[0]->mountInfo->getUser();
 
-		/** @var array<string, ACLManager> $userAclManagers */
-		$userAclManagers = [];
+		/** @var array<string, string> $relativePaths */
```
provokateurin:
```suggestion
		/** @var list<string> $relativePaths */
```

lib/Mount/MountProvider.php
```diff
@@ -287,24 +285,38 @@ public function getMountsForPath(string $setupPathHint, bool $forChildren, array
...
+		/** @var array<string, ACLManager> $userAclManagers */
+		$userAclManagers = [];
+
+		$folders = $this->folderManager->getFoldersForUser($user, null, $relativePaths);
+
+		foreach ($folders as $folder) {
+			$mountPoint = '/' . $user->getUID() . '/files/' . $folder->mountPoint;
+			if (isset($mounts[$mountPoint])) {
+				continue;
 			}
```
provokateurin:
```suggestion
```
This should not be necessary anymore, since all folders are queried at once and no possible duplicates are expected.

lib/Mount/MountProvider.php
```diff
@@ -287,24 +285,38 @@ public function getMountsForPath(string $setupPathHint, bool $forChildren, array
...
+
+			$userAclManagers[$user->getUID()] ??= $this->aclManagerFactory->getACLManager($user);
+			$aclManager = $userAclManagers[$user->getUID()];
```
provokateurin: The code above now assumes that we only every have one user (which is true AFAIK), so this can be simplified and moved outside the loop.

provokateurin (changes requested):
Thanks, I wanted to tackle this today!

lib/Mount/MountProvider.php
SystemKeeper: https://github.com/nextcloud/server/blob/e61f0074f02f67f6c5c25a6cbba06854a5111d24/lib/public/Files/Config/IPartialMountProvider.php#L39-L41

Yes, I already did it for `$user`, will do for ACL as well.

PR #4458, opened 2026-02-25
chore: Use non deprecated $appName variable

CarlSchwan:


lib/Controller/FolderController.php
```diff
@@ -42,7 +42,7 @@ class FolderController extends OCSController {
 	private readonly ?IUser $user;
 
 	public function __construct(
-		string $AppName,
+		string $appName,
```
susnux, line 45: or one could just use `Application::APP_ID`

PR #4487, opened 2026-03-05
refactor: Apply rector changes

provokateurin:


lib/ACL/RuleManager.php
```diff
@@ -36,9 +36,9 @@ private function createRule(array $data): ?Rule {
 		if ($mapping) {
 			return new Rule(
 				$mapping,
-				(int)$data['fileid'],
```
CarlSchwan, line 39: If this comes directly from the db, for oracle this needs to be explicitly casted to int

lib/Folder/FolderManager.php
```diff
@@ -159,7 +159,7 @@ public function getAllFoldersWithSize(int $offset = 0, ?int $limit = null, strin
 		/** @var list<array{folder_id: int, mount_point: string, quota: int, acl: bool, acl_default_no_permission: bool, storage_id: int, root_id: int, options: string}> $rows */
 		$rows = $query->executeQuery()->fetchAll();
 
-		$folderIds = array_map(static fn (array $row): int => (int)$row['folder_id'], $rows);
+		$folderIds = array_map(static fn (array $row): int => $row['folder_id'], $rows);
```
CarlSchwan: Same

lib/ACL/RuleManager.php
provokateurin: Figured this out too :/

CarlSchwan: * starts ranting about oracle :(

provokateurin:
Fixed the type annotations, so phpstan and rector will know that explicit casting is required.

PR #4500, opened 2026-03-11
fix(acls): allow uploading files with read+create+delete permissions

leftybournes:
This fixes an issue with accounts unable to upload files or folders without the edit permission. The logic was essentially copied from server's [permission mask wrapper](https://github.com/nextcloud/server/blob/1a70d192f3d328cd0e9ad74d5b1a0b45b5ab61e0/lib/private/Files/Storage/Wrapper/PermissionsMask.php#L65).

### DIFF
--- a/lib/ACL/ACLStorageWrapper.php
+++ b/lib/ACL/ACLStorageWrapper.php
@@ -85,12 +85,9 @@ class ACLStorageWrapper extends Wrapper implements IConstructableStorage {
 
 	#[\Override]
 	public function rename(string $source, string $target): bool {
-		if (str_starts_with($source, $target)) {
-			$part = substr($source, strlen($target));
-			//This is a rename of the transfer file to the original file
-			if (str_starts_with($part, '.ocTransferId')) {
-				return $this->checkPermissions($target, Constants::PERMISSION_CREATE) && parent::rename($source, $target);
-			}
+		// This is a rename of the transfer file to the original file
+		if (dirname($source) === dirname($target) && strpos($source, '.ocTransferId') > 0) {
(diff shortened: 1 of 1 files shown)


brlin-tw:
@leftybournes 

Hello, I'm also tackling the same issue and encountered this pull request as a result.

I would like to ask whether it would be more proper to verify whether the file's xxh128 hash match as well from the security perspective?

Something like:

```patch
--- a/apps/groupfolders/lib/ACL/ACLStorageWrapper.php
+++ b/apps/groupfolders/lib/ACL/ACLStorageWrapper.php
@@ -70,13 +70,15 @@
 	}
 
 	public function rename(string $source, string $target): bool {
-		if (str_starts_with($source, $target)) {
-			$part = substr($source, strlen($target));
-			//This is a rename of the transfer file to the original file
-			if (str_starts_with($part, '.ocTransferId')) {
+		//This is a rename of the transfer file to the original file
+		if (dirname($source) === dirname($target) && ($ocTransferIdPos = strpos(basename($source), '.ocTransferId')) > 0) {
+			$sourceHash = substr(basename($source), 0, $ocTransferIdPos);
+			$expectedHash = hash('xxh128', basename($target));
+			if ($sourceHash === $expectedHash) {
 				return $this->checkPermissions($target, Constants::PERMISSION_CREATE) && parent::rename($source, $target);
 			}
 		}
+
 		$permissions = $this->file_exists($target) ? Constants::PERMISSION_UPDATE : Constants::PERMISSION_CREATE;
 		$sourceParent = dirname($source);
 		if ($sourceParent === '.') {
```

Thanks for the fix, BTW.

PR #4516, opened 2026-03-25
fix(GroupFoldersHome): Show all leaf groupfolders

provokateurin:
If nested groupfolders are used they don't show up in the left sidebar.

### DIFF
--- a/lib/DAV/GroupFoldersHome.php
+++ b/lib/DAV/GroupFoldersHome.php
@@ -10,6 +10,7 @@ namespace OCA\GroupFolders\DAV;
 
 use OC\Files\Filesystem;
 use OCA\GroupFolders\Folder\FolderDefinition;
+use OCA\GroupFolders\Folder\FolderDefinitionWithPermissions;
 use OCA\GroupFolders\Folder\FolderManager;
 use OCP\Files\IRootFolder;
 use OCP\IUser;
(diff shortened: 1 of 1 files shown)


skjnldsv:
> If nested groupfolders are used they don't show up in the left sidebar.

wasn't it on purpose ?

provokateurin:
It's really confusing to users though. Especially if only use nested groupfolders and have none at the top, the list is just empty.

provokateurin:
After creating the PR I also noticed that I used "leaf" in the wrong way.
Nested groupfolders are basically disconnected trees starting at various depths within a bigger tree. Currently we only show the ones at depth 0, while this change ensures we show the top-most "root" groupfolder for any tree at any depth.
I hope this makes a bit more sense now.

lib/DAV/GroupFoldersHome.php
```diff
@@ -107,10 +108,18 @@ public function getChildren(): array {
 
 		$folders = $this->folderManager->getFoldersForUser($this->user);
 
-		// Filter out non top-level folders
-		$folders = array_filter($folders, fn (FolderDefinition $folder): bool => !str_contains($folder->mountPoint, '/'));
+		usort($folders, static fn (FolderDefinitionWithPermissions $a, FolderDefinitionWithPermissions $b): int => $a->mountPoint <=> $b->mountPoint);
```
skjnldsv, line 111: You can compare strings with a spaceship operator ? 🤯

provokateurin: It's the easiest way to do it in a sort function and we do it all the time :sweat_smile:

skjnldsv: Thought it was only for integers :see_no_evil:

provokateurin: PHP magic :crystal_ball:

Issue #4499, opened 2026-03-10
Team folders visible but not accessible after NC 33 upgrade


Crouze:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Upgrade from NC 32 to 33

### Expected behaviour
 As with NC 32, users can both see their group folders and open them.

### Actual behaviour
Users can still see the group folders, but trying to open these does not work. Either nothing happens, or "folder not found" is shown if pressing F5. Performing occ scans and repair have no effect. Group folders structure and data are still present.

### Server configuration

**Operating system:**
Arch Linux (up to date)

**Web server:**
Apache 2.4.66

**Database:**
Mariadb 12.2.2

**PHP version:**
PHP 8.5.3 (was 8.4.15 with NC 32)
PHP-legacy 8.3.30 (was 8.2.29 with NC 32)

**Nextcloud version:** (see Nextcloud admin page)
33.0.0 (was 32.0.5)

**Team folders version:**
21.0.6

**Updated from an older Nextcloud/ownCloud or fresh install:**
Upgraded from NC 32

**Where did you install Nextcloud from:**
Arch repository (Pacman)

**Are you using external storage, if yes which one:** local/s3/smb/sftp/...
cifs mount

**Are you using encryption:** yes/no
no

**Are you using an external user-backend, if yes which one:** LDAP/ActiveDirectory/Webdav/Saml/...
no

### Client configuration
**Browser:**
Firefox 148.0

**Operating system:**
Arch Linux (up-to-date)

provokateurin:
Hi, I'll take a look. I have an idea what could be causing this.

provokateurin:
I tried to reproduce with 21.0.6, but everything was working. Can you check your browser and server logs for any errors? Also check the network tab for failed requests and their responses.

Crouze:
I have in the mean time decided to move to a new instance of NC, this time docker'ized. Doing so I noticed a difference in file structure in __groupfolders between 32.0 and 33.0:

```
32.0:
---1---files---<files>
 |-2---files---<files>
 |-trash
 |-versions

33.0:
---1---files---<files>
     |-trash
     |-versions
---2---files---<files>
[...]

PR #4524, opened 2026-03-28
fix: Include fileid in route params

solracsf:
When clicking a folder in the Team Folders view, the `goToRoute` call omitted the `fileid` route param. Without it, the resolved route `/files?dir=/X` can collide with stale router state from a previous navigation, causing Vue Router to silently drop it as a `NavigationDuplicated`.

Adding `fileid` makes the target route `/files/123?dir=/X`, which is structurally distinct and navigates correctly.

On `stable32`:

https://github.com/nextcloud/groupfolders/blob/1ef5993faf35bb39869caa7844d8162f8e758783/src/actions/openGroupfolderAction.ts#L24

Returning `null cast to boolean` evaluates as falsy at runtime. The v3.x `FileAction` executor treated a falsy result as "action failed" and did not record this navigation as the successful result of a default action. This meant the router's state wasn't tied to this specific action's outcome, so the stale-route issue didn't surface before.

Fix https://github.com/nextcloud/groupfolders/issues/4499

### DIFF
--- a/src/actions/openGroupfolderAction.ts
+++ b/src/actions/openGroupfolderAction.ts
@@ -19,7 +19,7 @@ export const action: IFileAction = {
 		const dir = nodes[0].attributes.mountPoint
 		window.OCP.Files.Router.goToRoute(
 			null, // use default route
-			{ view: 'files' },
+			{ view: 'files', fileid: nodes[0].id },
 			{ dir },
 		)
 		return null


src/actions/openGroupfolderAction.ts
```diff
@@ -19,7 +19,7 @@ export const action: IFileAction = {
 		const dir = nodes[0].attributes.mountPoint
 		window.OCP.Files.Router.goToRoute(
 			null, // use default route
-			{ view: 'files' },
+			{ view: 'files', fileid: String(nodes[0].fileid) },
```
susnux: For 33+ its easier (and not deprecated) to use the snowflake-id aware `.id` attribute:


```suggestion
			{ view: 'files', fileid: nodes[0].id },
```

solracsf: Go for it :)

solracsf:
Fix for PHPstan: https://github.com/nextcloud/groupfolders/pull/4534

PR #4539, opened 2026-04-02
fix(quota): Invalidate root ETag on quota change

solracsf:
When  the quota of a team folder changes, desktop clients continued to receive the stale `quota-available-bytes` value on subsequent PROPFINDs until the `etag` of the folder root changed. This fix invalidates the root etag immediately after the quota is updated, without requiring a mounted storage or an active user session.

This forces every desktop client to issue a fresh PROPFIND on the next sync cycle, at which point the DAV layer computes and returns the updated `quota-available-bytes` value.

Server side: https://github.com/nextcloud/server/pull/59398
Needed for https://github.com/nextcloud/desktop/issues/4580


solracsf:
Adapted to use `Cache::update()`

Issue #4264, opened 2026-01-16
Creating new team folders does not work


SimJoSt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Go to the team folder admin interface
2. Type in the name of a new team folder and click "Create"

### Expected behaviour
The team folder should be created.

### Actual behaviour
Nothing happens in the interface, even when trying multiple times.
There are no errors shown in the interface, whatsoever.

### Server configuration

**Operating system:** Debian/Ubuntu

**Web server:** Apache

**Database:** MariaDB

**PHP version:** 8.3

**Nextcloud version:** 32.0.3

**Team folders version:** 20.1.7

**Updated from an older Nextcloud/ownCloud or fresh install:** Update

**Where did you install Nextcloud from:** Docker

**Are you using external storage, if yes which one:** no

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** no

**Operating system:** Debian/Ubuntu

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>
First error:

```php
[PHP] Warning: mkdir(): Permission denied at /var/www/html/custom_apps/groupfolders/lib/Mount/FolderStorageManager.php#130
	POST /apps/groupfolders/folders
	from ***REDACTED*** by ***REDACTED*** at 15.01.2026, 23:17:37
```

```json
{
  "reqId": "Eoli47AJe2AgZ9Li3IDE",
  "level": 2,
  "time": "2026-01-15T23:17:37+01:00",
  "remoteAddr": "***REDACTED***",
  "user": "***REDACTED***",
  "app": "PHP",
  "method": "POST",
  "url": "/apps/groupfolders/folders",
  "message": "mkdir(): Permission denied at /var/www/html/custom_apps/groupfolders/lib/Mount/FolderStorageManager.php#130",
  "userAgent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/143.0.0.0 Safari/537.36",
  "version": "32.0.3.2",
  "data": {
    "app": "PHP"
  },
[...]

PR #4552, opened 2026-04-10
fix: Add `datadirectory` fallback if not defined

solracsf:
Prevent things like https://github.com/nextcloud/groupfolders/issues/4264


provokateurin (approved):
Thanks!

PR #4586, opened 2026-04-21
fix(Action): await navigation and handle router rejections

solracsf:
PR [#4524](https://github.com/nextcloud/groupfolders/issues/4524) added fileid to the route params but left two async-contract bugs in the action itself: goToRoute wasn't awaited, and exec returned null (falsy, reported as failed to the v4 FileAction executor).

This PR awaits the navigation and returns true on success. Without this, the stale-route failure path in [#4499](https://github.com/nextcloud/groupfolders/issues/4499) remains observable on the top-level team-folder click path even once the server-side router fix ships.

Adds two E2E smoke cases asserting the observable behavior of clicking a team folder from the Team folders view.

### DIFF
--- a/src/actions/openGroupfolderAction.ts
+++ b/src/actions/openGroupfolderAction.ts
@@ -16,13 +16,23 @@ export const action: IFileAction = {
 	enabled: ({ view }) => view.id === appName,
 
 	async exec({ nodes }) {
-		const dir = nodes[0].attributes.mountPoint
-		window.OCP.Files.Router.goToRoute(
-			null, // use default route
-			{ view: 'files', fileid: nodes[0].id },
-			{ dir },
-		)
-		return null
+		try {
+			await window.OCP.Files.Router.goToRoute(
+				null, // use default route
+				{ view: 'files', fileid: nodes[0].id },
+				{ dir: nodes[0].attributes.mountPoint },
+			)
+			return true
+		} catch (e) {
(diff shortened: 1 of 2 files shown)


src/actions/openGroupfolderAction.ts
```diff
@@ -16,13 +16,23 @@ export const action: IFileAction = {
...
+				{ view: 'files', fileid: nodes[0].id },
+				{ dir: nodes[0].attributes.mountPoint },
+			)
+			return true
+		} catch (e) {
+			// Vue Router throws on duplicated/redirected navigations; those are not
+			// real failures from the user's perspective — the target view is reached.
+			const name = (e as { name?: string })?.name
+			const message = (e as { message?: string })?.message ?? ''
+			if (name === 'NavigationDuplicated' || /Redirected/.test(message)) {
```
susnux: Those should already be handled by file router in server (if not fixed there)

solracsf: `NavigationDuplicated` is handled at the server router layer (so my original catch for it was redundant), but `Redirected` is not handled anywhere, it bubbles as an uncaught promise rejection from every `goToRoute` call in every app. Neverthless, it should be handled there too.

solracsf: Follow-up: https://github.com/nextcloud/server/pull/59842

PR #4675, opened 2026-05-11
Add E2E test for searching with restricted ACL read permissions

danxuliu:
This is a E2E test for the issue in https://github.com/nextcloud/server/issues/60306. ~~Although the issue seems to be in server itself it looks like a good use case to have an explicit test for it in the _groupfolders_ app (and even if not at least provides an easy way to test the fix :-) ).~~ This will be fixed in the _groupfolders_ app itself in https://github.com/nextcloud/groupfolders/pull/4745 and its follow up https://github.com/nextcloud/groupfolders/pull/4817 Those pull requests already include a unit test for the issue, so this E2E test may or may not be needed. Up to the maintainers :-)

It is worth noting, though, that I noticed the regression in #4745 when I rebased this pull request and the test did not pass yet ;-)

~~It will need to be kept as a draft until the issue is fixed, but the test itself is ready for review.~~

### DIFF
--- a/cypress/e2e/groupfolders.cy.ts
+++ b/cypress/e2e/groupfolders.cy.ts
@@ -28,6 +28,12 @@ import {
 	PERMISSION_READ,
 	PERMISSION_WRITE,
 } from './groupfoldersUtils.ts'
+import {
+	currentSearchSectionCanLoadMoreResults,
+	currentSearchSectionHasResult,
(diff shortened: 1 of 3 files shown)


CarlSchwan (approved):

danxuliu:
Adjusted to the current behaviour of unified search.

PR #4682, opened 2026-05-13
fix: Show new folder creation on top and pagination sticky, fix basic design issues

### LINKED ISSUE #4465: [a11y] "new team folder" must not be a table row
### How to use GitHub

### Steps to reproduce
1. open admin settings
2. see that the "new team folder" ui is part of the table - this is semantically wrong

### Expected behaviour
Should not be part of existing team folders table

### Actual behaviour

Image

jancborchardt:
Just tried to keep the code changes low as the app doesn’t use Vue components. This fixes 2 specific issues as reported in our design channel:

> 1. Adding new team folders you have to scroll to the bottom of the list to reach folder creation field and button. Why can't it be in the top of the list?
> 2. Also the buttons to next page can only be reached when you first scrolled down to bottom. Please think about users with a larger number of rows.

Thanks to **KIVG** for that feedback. :) 

Before | After
-|-
Screenshot From 2026-05-13 15-44-33|Screenshot From 2026-05-13 15-43-39

### DIFF
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 #groupfolders-wrapper {
-	padding: 30px;
+	padding: calc(var(--default-grid-baseline) * 10) calc(var(--default-grid-baseline) * 10) calc(var(--default-grid-baseline) * 20);
 }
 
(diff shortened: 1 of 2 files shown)


kra-mo (approved):
Weird layout generally, but the change is nice

benjaminfrueh (approved):

Issue #2899, opened 2023-07-05
Add a UI for SSE for groupfolders


AndyScherzinger:
While the server supports SSE for groupfolders there currently isn't an admin UI for configuration of this, just CLI

PR #4698, opened 2026-05-23
feat: Implement UI for enabling Team Folders SSE in Nextcloud security settings

falghamdi125:
### ☑️ Resolves

- #2899 

### Summary

This PR adds a UI in the Nextcloud security settings to manage Server-Side Encryption (SSE) for Team Folders.

The goal is to make SSE for Team Folders more accessible and centralized from the security settings page instead of relying only on OCC method.

### Notes

- Enabling SSE for Team Folders from the UI is only possible when SSE is already enabled.
- Disabling SSE for Team Folders from the UI is intentionally not supported, similar to the existing global SSE enablement behavior.
- Disabling SSE for Team Folders remains possible through the OCC command.
- Translations have not yet been submitted to Transifex

Suggestions are welcome.

### DIFF
--- a/lib/Settings/AdminSecurity.php
+++ b/lib/Settings/AdminSecurity.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2026 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\GroupFolders\Settings;
+
+use OCA\GroupFolders\AppInfo\Application;
+use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\TemplateResponse;
+use OCP\AppFramework\Services\IInitialState;
+use OCP\IAppConfig;
+use OCP\Settings\ISettings;
+use OCP\Util;
+
+class AdminSecurity implements ISettings {
+	public function __construct(
+		private readonly IInitialState $initialState,
+		private readonly IAppConfig $appConfig,
+	) {
+	}
+
+	/**
+	 * @return TemplateResponse<Http::STATUS_OK, array{}>
+	 */
+	#[\Override]
+	public function getForm(): TemplateResponse {
+		$this->initialState->provideInitialState(
+			'server_side_encryption',
+			$this->appConfig->getValueString('core', 'encryption_enabled', 'no'),
+		);
+		$this->initialState->provideInitialState(
+			'enable_encryption',
+			$this->appConfig->getValueBool(Application::APP_ID, 'enable_encryption', false),
+		);
+
+		Util::addStyle(Application::APP_ID, Application::APP_ID . '-settings-security');
+		Util::addScript(Application::APP_ID, Application::APP_ID . '-settings-security');
+
+		return new TemplateResponse(Application::APP_ID, 'settings-security', []);
+	}
+
+	#[\Override]
+	public function getSection(): string {
+		return 'security';
+	}
+
+	#[\Override]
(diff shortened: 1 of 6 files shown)


come-nc (commented):
Can you provide a screenshot of the result?
Would it be possible to put this in the existing section rather than in a new one?
But if I understand the code correctly the current section is not migrated to vue yet, right?

falghamdi125:
Yes, it's possible to add the setting to the app section. However, I added the toggle section to the security settings because I thought it would be better for such an important option to be placed in the security section, so it can be enabled right after enabling SSE.

We could also move that option to the Groupfolders app section. Additionally, when the admin enables SSE, we can check whether the Groupfolders app is installed and prompt the admin e.g.  The Team folders app is installed. Do you want to enable SSE for team folders as well?

The current app section is written in React and not yet migrated to Vue.  

## Screenshots 

### Team folder encryption settings (No encryption module installed, SSE off)

image

### Team folder encryption settings (Encryption module installed, SSE off)

image

### Team folder encryption settings (SSE on, Team Folders encryption on)

image

src/components/TeamFoldersEncryption.vue
```diff
@@ -0,0 +1,74 @@
...
+
+async function handleChange(isEnabled: boolean) {
+	if (loading.value || serverSideEncryptionEnabled.value === 'no' || (folderEncryptionEnabled.value == true && isEnabled == false)) {
+		return
+	}	
+	const previous = folderEncryptionEnabled.value
+	folderEncryptionEnabled.value = isEnabled
+	loading.value = true
+	try {
+		await axios.post(generateUrl('apps/groupfolders/settings'), { enable: isEnabled })
```
artonge: I think you should be able to edit an app config with this endpoint: `/apps/provisioning_api/api/v1/config/apps`.
Example: https://github.com/nextcloud/server/blob/a2a1f525bfb1b5fb8d7bd003546ecfbd0e16a9cb/apps/settings/src/components/Encryption/EncryptionSettings.vue#L79

falghamdi125: The endpoint has been updated as suggested 👍

lib/Settings/AdminSecurity.php
```diff
@@ -0,0 +1,53 @@
...
+		$this->initialState->provideInitialState(
+			'server_side_encryption',
+			$this->appConfig->getValueString('core', 'encryption_enabled', 'no'),
+		);
+		$this->initialState->provideInitialState(
+			'enable_encryption',
+			$this->appConfig->getValueBool(Application::APP_ID, 'enable_encryption', false),
+		);
+
+		Util::addStyle('encryption', 'settings_admin');
```
artonge: Is this really needed?

falghamdi125: I think so. It provides the necessary styles (padding, margins, disable state icon, etc.) for the toggle so it looks and behaves exactly like the encryption toggles.

artonge: Can you double-check? Because we do not usually add resources of other applications, and naively, I would assume that the `groupfolders-settings-security.css` file should already contain all the necessary styling

falghamdi125: That line isn’t needed, as you suggested. It turned out the issue was related to my environment: I cleaned and upgraded the npm modules. The toggle now renders beautifully.

artonge (approved):

artonge:
@falghamdi125, your PR is looking good. The only blocking part is the psalm error that needs to be fixed :) 
The Cypress one we can ignore.

artonge:
Thank you for the contribution @falghamdi125 !

PR #4713, opened 2026-05-28
fix: properly distinguish unset ACL rules in UI

icewind1991:
Currently, the UI defaults to showing "Allow (Inherited permissions)" when no permission is set or inherited for a rule. This can lead to confusion since deny permissions from another rule might apply.

image


icewind1991:
It still needs a better way to distinguish "Allow"/"Deny" from "Allow (Inherit permissions)"/"Deny (Inherit permissions)" though.
In the above screenshot the first allow is inherited, while the second one is explicit, but you wouldn't know that without looking at the tooltip.

romeg0714 (approved):

provokateurin (approved):

PR #4714, opened 2026-05-28
fix: don't show inherited rule if there is a rule overwriting it

icewind1991:
Currently we always show all inherited ACLs as rules, even if there is an explicitly set rule on the same path that overwrites the inherited rule

Before:
image
(group "admin" has two rows)

After:
image

(Both screenshots with https://github.com/nextcloud/groupfolders/pull/4713 applied)


icewind1991:
Merged into https://github.com/nextcloud/groupfolders/pull/4729

MPStudyly:
~~From these screenshots it doesn't look clear which properties were inherited an which weren't. I don't think that's quite right. IMHO inherited properties should keep the gray coloring as long as not overwritten.~~

NVM, didn't see the other PRs.

PR #4730, opened 2026-06-02
show ACL group/user/team name in it's own row

icewind1991:
Trying to fit both the name and permission icons in the same row doesn't really work, especially with longer names.

Currently the name is only visible in the hover-text, which is bad for usability once you deal with more than one or two rules.

By giving the name it's own row there should be enough space for most names without cutting to much of.

Design can probably be tweaked a bit by someone who's better at css

Before:

image

After:

image

requires

 - [x] https://github.com/nextcloud/groupfolders/pull/4729

### DIFF
--- a/src/components/SharingSidebarView.vue
+++ b/src/components/SharingSidebarView.vue
@@ -324,10 +324,15 @@ async function changePermission(item: Rule, permission: number, state: number) {
 			</thead>
 			<tbody v-if="!aclCanManage">
 				<tr>
-					<td class="username">
+					<td>
 						<NcAvatar user="admin" :size="24" />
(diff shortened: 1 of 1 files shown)


provokateurin (commented):
While I understand the problem, I think it makes the view less usable due to the loss of density.

provokateurin (approved):

marcoambrosini:
Hi @icewind1991, this breaks the table layout by inserting a row and makes readability quite hard. While we think about a more permanent solution, could we go back to the original state with the ellipsized team name? 
What were the exact reasons for the change to the avatar-only view?

PR #4736, opened 2026-06-03
perf(folder): avoid on-disk temp table when sorting folders by group count

solracsf:
## Summary

Found while working on https://github.com/nextcloud/server/pull/60932

The admin list, when sorted by the **Groups** column, calls `FolderManager::getAllFoldersWithSize($offset, $limit, 'groups', $order)`. That code path ordered folders by their applicable-group count with a `LEFT JOIN group_folders_groups + GROUP BY f.folder_id`:

```php
$query
    ->leftJoin('f', 'group_folders_groups', 'g', $query->expr()->eq('f.folder_id', 'g.folder_id'))
    ->groupBy('f.folder_id')
    ->orderBy($query->func()->count('g.applicable_id'), $order);
```

The `GROUP BY` forces the database to materialize an internal temporary table. Because `selectWithFileCache()` selects the `options` column, a `LONGTEXT`, that temporary table cannot use the in-memory `MEMORY` engine (it cannot hold TEXT/BLOB), so MariaDB/MySQL **always** spill it to an on-disk (Aria) temporary table, regardless of `tmp_table_size`, on every request that uses this sort.

### Validation (MariaDB 10.11, 3 group folders, 60k `oc_filecache` rows)

`EXPLAIN`, driving table `f`:

| | Extra |
|---|---|
| before | `Using where; Using temporary; Using filesort` |
| after | `Using where; Using filesort` |

`SHOW STATUS` around the query (`tmp_table_size = max_heap_table_size = 64M`):

| | `Created_tmp_disk_tables` |
|---|---|
| before | **1** |
| after | **0** |

Result-equivalence: the returned folder order is identical for both `ASC` and `DESC`, verified by building the query through `IQueryBuilder` against the real database.

## Tests

Adds `FolderManagerTest::testGetAllFoldersWithSizeOrderedByGroups`, which creates folders with 1/2/3 applicable groups and asserts the `'groups'` sort returns them ordered by group count (both ascending and descending).

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -159,10 +159,20 @@ class FolderManager {
 		$query->setFirstResult($offset);
 		$query->setMaxResults($limit);
 		if ($orderBy === 'groups') {
-			$query
-				->leftJoin('f', 'group_folders_groups', 'g', $query->expr()->eq('f.folder_id', 'g.folder_id'))
-				->groupBy('f.folder_id')
-				->orderBy($query->func()->count('g.applicable_id'), $order);
+			// Order by the number of groups/circles applicable to each folder.
+			//
+			// We deliberately avoid a "JOIN group_folders_groups + GROUP BY f.folder_id"
+			// here. Grouping forces the database to materialize an internal temporary
+			// table, and because the selected `options` column is a LONGTEXT that
+			// temporary table cannot use the in-memory MEMORY engine, so MariaDB/MySQL
+			// always spills it to an on-disk (Aria) temp table regardless of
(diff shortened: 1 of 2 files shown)


icewind1991:
It's probably best to just drop the ordering altogether

solracsf:
Cc @provokateurin since this has been introduced by https://github.com/nextcloud/groupfolders/commit/38dfb303c0aedc14ef4396e7ce173e33f3af0e23

CarlSchwan:
Alternative https://github.com/nextcloud/groupfolders/pull/4863

PR #4737, opened 2026-06-03
fix(acl): make preloadRulesForFolder actually warm the rule cache

solracsf:
## Summary

`ACLManager::preloadRulesForFolder()` was a **no-op**: it ran a query that scans every child of a folder and then threw the result away, caching nothing. This fixes it to populate the rule cache as intended, and speeds up the underlying query.

`preloadRulesForFolder()` called `RuleManager::getRulesForFilesByParent()` but ignored the returned array:

```php
public function preloadRulesForFolder(int $storageId, int $parentId): void {
    $this->ruleManager->getRulesForFilesByParent($this->user, $storageId, $parentId);
}
```

`getRulesForFilesByParent()` has no caching side effect: it just builds and returns rules. So on every trash listing (`TrashBackend::listTrashFolder()` and `getTrashForFolders()`), this query scanned every child of the folder (~20k filecache rows on a large folder) and cached nothing. The intent was clearly to warm `ACLManager::$ruleCache` so the subsequent per-child `getACLPermissionsForPath()` calls would be cache hits.

Measured ~**3.6x faster** for the preload query on a 15k-ACL / 20k-child folder (MariaDB 10.11), plus the per-child lookups for children that have rules are now served from cache instead of issuing a query per path.

### DIFF
--- a/lib/ACL/ACLManager.php
+++ b/lib/ACL/ACLManager.php
@@ -241,8 +241,18 @@ class ACLManager {
 		}
 	}
 
+	/**
+	 * Warm the rule cache for the direct children of a folder.
+	 *
+	 * Used before listing a folder with many children (e.g. trash listings),
+	 * so the subsequent per-child `getACLPermissionsForPath` lookups are
+	 * served from the cache instead of querying the rules for each child path.
+	 */
 	public function preloadRulesForFolder(int $storageId, int $parentId): void {
-		$this->ruleManager->getRulesForFilesByParent($this->user, $storageId, $parentId);
+		$rules = $this->ruleManager->getRulesForFilesByParent($this->user, $storageId, $parentId);
+		foreach ($rules as $path => $rulesForPath) {
+			$this->ruleCache->set($path, $rulesForPath);
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


icewind1991:
The `left join` is intentional here and necessary to ensure that paths without rules also get cached.

This broke because of https://github.com/nextcloud/groupfolders/pull/3942 though, but that's an easy fix

icewind1991:
https://github.com/nextcloud/groupfolders/pull/4738

solracsf:
@icewind1991 IMO #4738 needs the ACLManager change in this PR to actually achieve "files without rules also get cached."

Basically:

- Revert the RuleManager.php change entirely in this PR (drop the INNER JOIN and the loop simplification) as #4738 owns that file and does it correctly
- Keep the `ACLManager::preloadRulesForFolder` cache-population loop + the `ACLManagerTest` test (the test mocks `getRulesForFilesByParent`, so it's unaffected by the revert).

Can you check/confirm?

icewind1991:
Yes, this PR is still needed

solracsf:
Reverted the RuleManager.php change entirely in this PR

icewind1991 (approved):

PR #4738, opened 2026-06-03
return results for files without rules in RuleManager

icewind1991:
Empty results are important to ensure that files without rules also get cached.

This broke with https://github.com/nextcloud/groupfolders/pull/3942


solracsf:
See https://github.com/nextcloud/groupfolders/pull/4737#issuecomment-4612759289

provokateurin (approved):

PR #4744, opened 2026-06-04
perf(acl): share one ACLManager per user per request

### LINKED ISSUE #4740: ACL rule cache: preload is orphaned because a new `ACLManager` (and a new cache) is created per call
The issue:

`ACLManager` memoizes resolved ACL rules in a per-instance `CappedMemoryCache` (`$ruleCache`). `TrashBackend` tries to warm that cache before a trash listing via `preloadRulesForFolder()`. But `ACLManagerFactory::getACLManager()` returns a **fresh `ACLManager` (with a fresh empty cache) on every call**, so the instance that gets preloaded is discarded, and the per-item permission checks run on *different* instances that never see the warmed cache.

Net effect: the trash-listing ACL preload provides **no measurable benefit**, and there is **no cache reuse across trash items** at all. This is independent of (and compounds with) the two rule-cache bugs being fixed in #4737 (preload discarded its result) and #4738 (#3942 stopped returning empty entries), even with both of those merged, the preload remains orphaned because of the per-call instantiation described here.

### Details

`ACLManagerFactory::getACLManager()` (`lib/ACL/ACLManagerFactory.php`) always constructs a new instance:

```php
public function getACLManager(IUser $user): ACLManager {
    return new ACLManager(/* ... */);   // fresh instance every call
}
```

Each `ACLManager` owns its own cache (`lib/ACL/ACLManager.php`):

```php
public function __construct(/* ... */) {
    $this->ruleCache = new CappedMemoryCache();  // per-instance, empty
}
```

### Impact

- Trash listings of group folders with many children re-resolve ACL rules per item (two `getACLPermissionsForPath()` calls each), defeating the intended preload.
- Folders with ACL enabled and large/active trash are the worst case.
- Correctness is unaffected, this is purely a performance/optimization regression.

solracsf:
Fix https://github.com/nextcloud/groupfolders/issues/4740

`ACLManagerFactory::getACLManager()` created a fresh ACLManager (and a fresh empty rule cache) on every call, so `preloadRulesForFolder()` warmed a cache that was immediately discarded and the per-item permission checks ran on different instances that never saw it.

Memoize the manager by UID. Since the factory is request-scoped, all callers now share one rule cache per user, so cache warming and the subsequent lookups actually hit the same cache.

Because the cache is now shared across all of a user's folders, scope its keys by storage id. Paths are only unique within a storage (folders using a separate storage start over at "`files/...`"), so a path-only key could
otherwise return one storage's rules for a same-named path in another.

Tests:

- `ACLManagerFactoryTest`: same UID returns the same instance; different UIDs return different instances.
- `ACLManagerTest::testRuleCacheIsScopedPerStorage`: the same relative path under two storage ids resolves with each storage's own rules (no cross-storage cache collision).

### DIFF
--- a/lib/ACL/ACLManagerFactory.php
+++ b/lib/ACL/ACLManagerFactory.php
@@ -9,23 +9,48 @@ declare(strict_types=1);
 namespace OCA\GroupFolders\ACL;
 
 use OCA\GroupFolders\ACL\UserMapping\IUserMappingManager;
+use OCP\Cache\CappedMemoryCache;
 use OCP\IAppConfig;
 use OCP\IUser;
 
 class ACLManagerFactory {
+	/**
+	 * Memoize one ACLManager (and therefore one rule cache) per user for the
+	 * lifetime of the request. The factory is a request-scoped service, so
+	 * sharing the instance lets cache warming (e.g. preloadRulesForFolder) and
+	 * the subsequent per-item permission checks hit the same cache instead of
+	 * each call resolving the rules again. ACL and group changes take effect on
+	 * the next request.
+	 *
+	 * Capped so a request resolving many users (e.g. an occ command or background
+	 * job) cannot grow the map without bound.
+	 *
+	 * @var CappedMemoryCache<ACLManager>
+	 */
+	private readonly CappedMemoryCache $managers;
+
 	public function __construct(
 		private readonly RuleManager $ruleManager,
 		private readonly IAppConfig $config,
 		private readonly IUserMappingManager $userMappingManager,
 	) {
+		$this->managers = new CappedMemoryCache();
 	}
 
 	public function getACLManager(IUser $user): ACLManager {
-		return new ACLManager(
-			$this->ruleManager,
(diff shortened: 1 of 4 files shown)


lib/ACL/ACLManagerFactory.php
```diff
@@ -13,6 +13,18 @@
...
+	 * Memoize one ACLManager (and therefore one rule cache) per user for the
+	 * lifetime of the request. The factory is a request-scoped service, so
+	 * sharing the instance lets cache warming (e.g. preloadRulesForFolder) and
+	 * the subsequent per-item permission checks hit the same cache instead of
+	 * each call resolving the rules again. ACL and group changes take effect on
+	 * the next request.
+	 *
+	 * @var array<string, ACLManager>
+	 */
+	private array $managers = [];
```
CarlSchwan: Use a CappedMemoryCache instead

solracsf: Done

icewind1991 (approved):

PR #4745, opened 2026-06-04
filter ACL forbidden paths from search query

icewind1991:
Currently, the logic for removing files the user doesn't have access to from search is done as a post-processing step.

While this filters out the correct entries, it means that the number of returned results is less than the limit asked for (or even 0). Which the search logic interprets as there being no more results.

By applying the filtering as part of the query, we return the number of results asked for.

### DIFF
--- a/lib/ACL/ACLCacheWrapper.php
+++ b/lib/ACL/ACLCacheWrapper.php
@@ -9,9 +9,14 @@ declare(strict_types=1);
 namespace OCA\GroupFolders\ACL;
 
 use OC\Files\Cache\Wrapper\CacheWrapper;
+use OC\Files\Search\SearchBinaryOperator;
+use OC\Files\Search\SearchComparison;
 use OCP\Constants;
 use OCP\Files\Cache\ICache;
 use OCP\Files\Cache\ICacheEntry;
+use OCP\Files\Search\ISearchBinaryOperator;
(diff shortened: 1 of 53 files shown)


icewind1991:
Fixes https://github.com/nextcloud/server/issues/60306

provokateurin (approved):

danxuliu:
@icewind1991 @provokateurin While the fix solves the issue for users with forbidden paths it breaks searching for users without any forbidden path. It can be reproduced with the steps from https://github.com/nextcloud/server/issues/60306 but searching as the user _manager_ instead of as the user _restricted_. The E2E test from https://github.com/nextcloud/groupfolders/pull/4675 also fails for the search done as user _manager_.

Adding something like:
```
		if (empty($forbiddenPaths)) {
			return new SearchBinaryOperator(ISearchBinaryOperator::OPERATOR_AND, []);
		}
```
to [`getSearchFilter()`](https://github.com/nextcloud/groupfolders/pull/4745/changes/98ce44b3c31c2723401df476cc31352d96d52df3#diff-915d11479fd233c02aca5c29e469c319a704058a89ca0d89e81adcd1921b26feR120) make it work as expected for both _manager_ and _restricted_, but I did not dig into the issue.

Besides that, would it be possible to backport this pull request and the follow up also to _stable32_? Thanks!

PR #4817, opened 2026-06-19
fix: fix search results when no paths are forbidden

icewind1991:
I'm not sure why `not in []` doesn't just become a noop.

## 🤖 AI (if applicable)


danxuliu (approved):
Tested and works 👍

Nevertheless, maybe it would be better to use `if (count($forbiddenPaths) === 0) {` or `if ($forbiddenPaths === []) {` instead of the current `if (empty($forbiddenPaths)) {`? (which [I suggested myself](https://github.com/nextcloud/groupfolders/pull/4745#issuecomment-4723055838) 🤦, but I always forget that in PHP [`empty()` is not only about arrays](https://www.php.net/empty))

PR #4825, opened 2026-06-21
fix(a11y): make team folder edit buttons keyboard-operable

### LINKED ISSUE #4466: [a11y] edit buttons in team folder table not interactive with keyboard
### How to use GitHub

### Steps to reproduce
1. open admin settings
2. create new team folder
3. try to edit the name with only your keyboard

### Expected behaviour
Can focus the "edit name" button with keyboard

### Actual behaviour
Button only visible on mouse-hover.

Image

mvanhorn:
## Summary

The team folders admin settings table rendered its row action controls as bare `<a>` elements with an `onClick` handler but no `href`:

- the rename trigger (the mount-point name), and
- the delete control (`<a className="icon icon-delete icon-visible">`).

Anchors without an `href` are not part of the tab order and do not respond to Enter/Space activation, so keyboard-only and screen-reader users could not rename or delete a team folder from the settings table.

This PR replaces those non-interactive anchors with native `<button type="button">` elements that carry the existing `onClick` handlers and a discernible `aria-label`. Buttons are natively focusable and fire on Enter/Space, restoring keyboard operability without a custom keydown handler. Minimal SCSS strips the default button chrome so the controls keep their prior icon/text appearance.

Notable accessibility detail: the rename button's `aria-label` includes the folder name (`Rename "{mountPoint}"`) so each row remains identifiable to screen-reader and voice-control users rather than every row being announced only as "Rename". The button-chrome reset is scoped to `button.action-rename` so the sibling `<a class="action-rename">` in the groups column keeps its existing link styling.

## Why this matters

Closes a WCAG keyboard-operability gap (2.1.1 Keyboard) in the team folders settings UI. The issue was filed by a maintainer and triaged by the Nextcloud team to the actionable `1. to develop` state (ref #4466). Scope is intentionally limited to the keyboard-interactivity fix for the row action controls.

## Testing

- Keyboard: Tab now reaches each row's rename and delete control, and Enter/Space activates the same handler the click path invokes (rename enters edit mode; delete prompts/removes the folder).
- Screen reader: each control exposes a non-empty accessible name ("Rename \"<folder>\"" / "Delete") rather than an empty, non-focusable anchor.
- Visual regression: the icon-only delete button and the inline rename control render with the same size/position/glyph as the prior anchors (no button background, border, or extra padding).
- Mouse: clicking the controls still renames/deletes exactly as before.
- `eslint src/settings/App.tsx` and `stylelint src/settings/App.scss` report no new findings on the changed lines (pre-existing warnings/errors unrelated to this change remain).

Fixes #4466

### DIFF
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -44,9 +44,11 @@
 			}
 		}
 
-		.action-rename {
-			// Reset default button chrome so the focusable rename control
-			// keeps the inline text appearance of the former <a> element.
+		// Reset default button chrome so the focusable rename control keeps
+		// the inline text appearance of the former <a>. Scoped to the button
+		// so the sibling <a class="action-rename"> in the groups column keeps
+		// its own link styling.
+		button.action-rename {
 			display: inline;
 			padding: 0;
 			margin: 0;
@@ -56,7 +58,9 @@
 			color: inherit;
 			text-align: inherit;
 			cursor: pointer;
+		}
 
+		.action-rename {
 			&::after {
 				display: inline-block;
 				position: relative;
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -192,7 +192,8 @@
 					color: var(--color-text-lighter);
 				}
 				&.mountpoint {
-					a {
+					a,
+					button.action-rename {
 						color: var(--color-main-text);
 						font-weight: bold;
 					}
--- a/src/settings/App.scss
+++ b/src/settings/App.scss
@@ -45,6 +45,18 @@
 		}
 
 		.action-rename {
+			// Reset default button chrome so the focusable rename control
+			// keeps the inline text appearance of the former <a> element.
+			display: inline;
+			padding: 0;
+			margin: 0;
+			border: none;
+			background: none;
+			font: inherit;
+			color: inherit;
+			text-align: inherit;
+			cursor: pointer;
+
 			&::after {
 				display: inline-block;
 				position: relative;
@@ -87,7 +99,8 @@
 			width: 32px;
 		}
 
-		a.icon {
+		a.icon,
+		button.icon {
 			margin-left: 5px;
 		}
 	}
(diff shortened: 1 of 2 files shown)


solracsf:
The rename control now renders as `<button class="action-rename">`, which no longer matches `td.mountpoint a`, so `font-weight: bold` is never applied. The new reset rule (`button.action-rename { font: inherit; color: inherit; }`) then makes the button inherit `font-weight: normal` and the ambient text color from `td.mountpoint`:

```
td {
    a { color: var(--color-text-lighter); }
    &.mountpoint {
        a { color: var(--color-main-text); font-weight: bold; }
    }
}
```

Failure scenario: in the admin team-folders table the folder name in each row, previously rendered bold in `--color-main-text`, now renders at normal weight, a visible regression that directly contradicts the PR's "renders with the same size/position/glyph as the prior anchors" claim. Fix: add `font-weight: bold;` (and `color: var(--color-main-text);` if you want to be exact) to the `button.action-rename` reset, or broaden the `td.mountpoint` selector to include `button.action-rename`.

mvanhorn:
Good catch, thanks. The mountpoint name moved from `<a>` to `<button class="action-rename">`, so `td.mountpoint a` stopped matching and it lost the bold/main-text emphasis. Fixed in c27c628 by extending that rule to `button.action-rename` - its higher specificity overrides the `color: inherit` / `font: inherit` from the button-chrome reset, so the folder name renders bold again. Verified stylelint clean and the vite production build passes.

susnux (approved):

susnux:
Thank you for your pull request, you need to sign-off your commits.
You can do so with `git commit -s` or to fix it now: `git commit --amend -s`

mvanhorn:
Thanks for the catch - signed off the commits (the last one was missing its Signed-off-by). The DCO check is green now.

mvanhorn:
Thanks for picking this up, @AndyScherzinger - keyboard-operable edit buttons close out the team folder a11y pass.

AndyScherzinger:
Well, thank you for caring and taking the time contribute a fix. Highly appreciated and welcomed 🎉

PR #4833, opened 2026-06-29
fix(settings): Do not initiate ACL search right away

provokateurin:
Each Team folder row with ACL enabled was issuing a request to the server when the settings page was loaded.


ShGKme (approved):
Makes sense for the initial state.

However, do I understand it correctly that it still will load (and may die on overloading) on any search?

ShGKme (approved):
I missed that this is for every row. Makes total sense now. Double approve.

PR #4852, opened 2026-07-01
perf: Add per-request caching for user mapping

CarlSchwan:
## 🤖 AI (if applicable)

### DIFF
--- a/lib/ACL/UserMapping/UserMappingManager.php
+++ b/lib/ACL/UserMapping/UserMappingManager.php
@@ -14,6 +14,7 @@ use OCA\Circles\Exceptions\CircleNotFoundException;
 use OCA\Circles\Model\Circle;
 use OCA\Circles\Model\Probes\CircleProbe;
 use OCP\AutoloadNotAllowedException;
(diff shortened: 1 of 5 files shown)


provokateurin (approved):

lib/ACL/UserMapping/UserMappingManager.php
```diff
@@ -57,10 +85,13 @@ public function mappingFromId(string $type, string $id): ?IUserMapping {
 		}
 
 		if ($displayName === null) {
+			$this->mappingByKey->set($cacheKey, false);
```
solracsf, line 88: Is the negative caching (false) intentional here?

It's per-request so it can't go stale across requests, but it can within a single long-running process (occ/cron): if a group/user/circle is looked up before it exists and created later in the same run, the cached false sticks and `RuleManager::createRule()` may silently drop that rule.

CarlSchwan: Good point, I now reset this cache between invocation of background jobs, just to be extra sure. For occ commands, I think the group membership when the command was started is good enough and might even be better

PR #4854, opened 2026-07-01
perf: Optimize call to userInMappings when no circle mapping is present

CarlSchwan:
Then we just need to load the list of group ids the user is in and not additionally the group display name which is the expensive part.

## 🤖 AI (if applicable)


susnux (approved):
looks good

PR #4863, opened 2026-07-02
perf: Make oc_group_folders.options a VARCHAR(255)

CarlSchwan:
Allow to query this column without on-disk temp table on MySQL

## 🤖 AI (if applicable)

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -21,7 +21,7 @@ As of Hub 10 / Nextcloud 31, admins must be members of a team to assign that tea
 Folder.
 ]]>
 	</description>
-	<version>23.0.0-dev.0</version>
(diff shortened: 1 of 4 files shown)


come-nc (approved):
Will the value always fit?

solracsf:
@come-nc, yes, as far as I understand it.

Reasoning: the column is typed `array{separate-storage?: bool}` and has only that single boolean key since it was introduced. `createFolder` is the sole writer. If someone adds enough option keys to exceed 255 later, strict mode (default since MySQL 5.7 / MariaDB 10.2.4) fails the insert loudly, so it forces a column widening rather than corrupting silently.

solracsf (approved):

PR #4874, opened 2026-07-06
perf: Cache whether an user is a groupfolder subadmin

CarlSchwan:
AuthorizedGroupMapper::findAllClassesForUser is expensive as the list of groups is computed each time.

## 🤖 AI (if applicable)

### DIFF
--- a/lib/Folder/FolderManager.php
+++ b/lib/Folder/FolderManager.php
@@ -74,6 +74,9 @@ class FolderManager {
 	 */
 	private array $canManageACLCache = [];
 
+	/** @var array<string, bool> $isSubAdminCache */
+	private array $isSubAdminCache = [];
+
 	public function __construct(
 		private readonly IDBConnection $connection,
 		private readonly IGroupManager $groupManager,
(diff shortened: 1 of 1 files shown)


lib/Folder/FolderManager.php
```diff
@@ -510,10 +513,14 @@ private function computeCanManageACL(int $folderId, IUser $user, bool $excludeAd
 		}
 
 		// Call private server api
-		if (class_exists(\OC\Settings\AuthorizedGroupMapper::class)) {
+		if (!$excludeAdmins && class_exists(\OC\Settings\AuthorizedGroupMapper::class)) {
 			$authorizedGroupMapper = Server::get(\OC\Settings\AuthorizedGroupMapper::class);
```
solracsf: `Server::get()` still runs on cache hits and is fetched on every call but only used inside the `!isset()` branch. Move it inside that branch so cache hits skip the container lookup :)

solracsf (approved):

solracsf:
Worth backporting it?

CarlSchwan:
> Worth backporting it?

Btw I have meanwhile also optimized the codepath in server, so this patch is less useful

PR #4875, opened 2026-07-08
perf: Don't fetch all the trash items when deleting a single item

CarlSchwan:
## 🤖 AI (if applicable)

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -438,7 +438,6 @@ class TrashBackend implements ITrashBackend {
 			// we apply acl filtering later to get the correct permissions again
 			$trashFolder = $this->setupTrashFolder($folder);
 			$content = $trashFolder->getDirectoryListing();
-			$userCanManageAcl = $this->folderManager->canManageACL($folder->id, $user, true);
 
 			$itemsForFolder = array_map(function (Node $item) use ($user, $folder, $indexedRows): \OCA\GroupFolders\Trash\GroupTrashItem {
 				$pathParts = pathinfo($item->getName());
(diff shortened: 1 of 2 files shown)


lib/Trash/TrashBackend.php
```diff
@@ -488,6 +494,79 @@ private function getTrashForFolders(IUser $user, array $folders): array {
...
+
+		foreach ($folders as $folder) {
+			// note that we explicitly don't pass the user here, was we need to get all trash items,
+			// not only the trash items we have access to (so we can get their original paths)
+			// we apply acl filtering later to get the correct permissions again
+			$trashFolder = $this->setupTrashFolder($folder);
+			try {
+				$item = $trashFolder->get($name);
+			} catch (NotFoundException) {
+				continue;
```
provokateurin, line 521: Would it be possible to figure out the correct folder for an item name without iterating all folders? If that isn't possible, then iterating the folder is a problem if a name is not unique to all folders (which could result in loading an item from the wrong folder).

provokateurin (changes requested):

lib/Trash/TrashBackend.php
CarlSchwan: I need to go back to the thinking phase :(

CarlSchwan: So tested a bit more and the current code works fine in this situation. This is because $name include the deletion timestamp so the risk of collision is very small.

The code in master has the same collision issue as if two files with the same name are deleted in the same second, we returns two items and delete the first one

provokateurin (approved):

PR #4928, opened 2026-07-20
fix: Move trash items cross-storage when updating trashed children

CarlSchwan:
Otherwise the group_folders_trash entries will be updated with the new folder_id but not the filecache entries.

This is also relevant when using object stores and a bucket per groupfolder.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly generated using AI (the new unit test in tests/Trash/TrashBackendTest.php, rest was by an human)

### DIFF
--- a/lib/Trash/TrashBackend.php
+++ b/lib/Trash/TrashBackend.php
@@ -718,4 +718,63 @@ class TrashBackend implements ITrashBackend {
 			$this->logger->error($e->getMessage(), ['exception' => $e]);
 		}
 	}
+
+	/**
+	 * Find trash items whose `group_folders_trash` row was updated to point to a
+	 * new Team folder but the actual entries weren't.
+	 *
+	 * @return int the number of trash items that were moved
+	 */
+	public function repairMisplacedTrashItems(): int {
+		$folders = $this->folderManager->getAllFolders();
+		if ($folders === []) {
+			return 0;
+		}
+
+		$trashFoldersById = [];
+		$nodesByFilename = [];
+		foreach ($folders as $folderId => $folder) {
+			try {
+				$trashFoldersById[$folderId] = $trashFolder = $this->setupTrashFolder($folder);
+			} catch (\Exception $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+				continue;
+			}
+
+			foreach ($trashFolder->getDirectoryListing() as $node) {
+				$nodesByFilename[$node->getName()][] = ['folderId' => $folderId, 'node' => $node];
+			}
+		}
+
+		$rows = $this->trashManager->listTrashForFolders(array_keys($folders));
+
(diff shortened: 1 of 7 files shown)


lib/Trash/TrashBackend.php
```diff
@@ -679,4 +679,102 @@ public function expire(Expiration $expiration): array {
...
+		foreach ($folders as $folderId => $folder) {
+			try {
+				$trashFoldersById[$folderId] = $trashFolder = $this->setupTrashFolder($folder);
+			} catch (\Exception $e) {
+				$this->logger->error($e->getMessage(), ['exception' => $e]);
+				continue;
+			}
+
+			foreach ($trashFolder->getDirectoryListing() as $node) {
+				$nodesByFilename[$node->getName()][] = ['folderId' => $folderId, 'node' => $node];
```
artonge: Could two files with the same name but in different groupfolder conflict?

```
_groupfolder1
 |__myfile1.txt
_groupfolder2
 |__myfile1.txt 
``` 

Same question with sub folder inside the same groupfolder.

```
_groupfolder1
 |__myfolder1
    |__myfile1.txt
 |__myfolder2
    |__myfile1.txt
```

lib/Trash/TrashBackend.php
```diff
@@ -679,4 +679,102 @@ public function expire(Expiration $expiration): array {
...
+	 * @return int the number of trash items that were moved
+	 */
+	public function repairMisplacedTrashItems(): int {
+		$folders = $this->folderManager->getAllFolders();
+		if ($folders === []) {
+			return 0;
+		}
+
+		$trashFoldersById = [];
+		$nodesByFilename = [];
```
artonge: Will this keep all the trashed files from groupfolders in memory at the same time?

icewind1991 (approved):

lib/Trash/TrashBackend.php
CarlSchwan: Only if they have been deleted at exactly the same time since the name of files and folders in the trash are always suffixed by `d.timetstamp`

lib/Trash/TrashBackend.php
CarlSchwan: Yes :( This should not be a huge problem since the upgrade for big instance is done via CLI

artonge: How much is the memory impact of a single node? IIRC we had users having trouble simply opening their trashbin due to its size, I can only imagine that for large customer, having all trashbin nodes in memory would explode.

Is it really needed to have them all in memory? Can't we do it node by node? Or at least groupfolder by groupfolder?

CarlSchwan: I'll try to come up with a solution

CarlSchwan: done :)

PR #4941, opened 2026-07-24
feat: add event handling to auto create teamfolders upon circle creation

stediefan:
feat: add team_circle_id flag to teamfolder to uniquely link a team
feat: adminUI: add filters for classic teamfolders and the newly linked ones
todo: sensible naming for the two types of folders

Assisted by: GitHub Copilot:Claude Sonnet 5
Assisted by: GitHub Copilot:GPT-5.6 Terra
Assisted by: GitHub Copilot:Kimi K2.7 Code

Related Issues in Circles:
https://github.com/nextcloud/circles/issues/2661
https://github.com/nextcloud/circles/issues/2663
https://github.com/nextcloud/circles/issues/2669

Related Pull Request in server: https://github.com/nextcloud/server/pull/62476
Related Pull Request in circles: https://github.com/nextcloud/circles/pull/2675

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI


lib/Command/ACL.php
```diff
@@ -66,9 +66,19 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		}
 
 		if ($input->getOption('enable')) {
-			$this->folderManager->setFolderACL($folder->id, true);
+			try {
```
CarlSchwan: I don't think these try catch are really needed. Same with the other commands

lib/Migration/Version2300000Date20260720000001.php
```diff
@@ -0,0 +1,136 @@
...
+				[
+					'notnull' => false,
+					'length' => 31,
+					'default' => null,
+				]
+			);
+		}
+
+		if (!$table->hasIndex('group_folders_team_circle')) {
+			$table->addUniqueIndex(['team_circle_id'], 'group_folders_team_circle');
```
CarlSchwan, line 48: as far as I can see, this wasn't enforced previously with the options.circle_id, so this might cause some issues

tests/stubs/ocp_teams_ifolderprovider.php
```diff
@@ -0,0 +1,15 @@
+<?php
```
CarlSchwan: This won't be needed as soon as the PR in server is merged

lib/Migration/Version2300000Date20260720000001.php
stediefan: There was some additional code I overlooked. Should be more concise now: 
This index belongs to a new property that links the folder 1:1 to a team. As the property is newly created we shouldn't have any issues here.

lib/Command/ACL.php
stediefan: Thx, Done.

tests/stubs/ocp_teams_ifolderprovider.php
stediefan: removed 👍

lib/TeamSpace/TeamSpaceProvider.php
```diff
@@ -0,0 +1,101 @@
...
+			return [];
+		}
+
+		return [new TeamResource(
+			$this,
+			(string)$folder->getId(),
+			$folder->getMountPoint(),
+			$this->urlGenerator->getAbsoluteURL('/apps/files/?dir=/' . rawurlencode($folder->getMountPoint())),
+			iconSvg: $this->getIconSvg(),
+		)];
```
artonge: Could we make it so that all groupfolders that are accessible to a team is returned in that method? This way, non-teamspace goupfolders would also be displayed as team resources.

stediefan: Added this function in the latest commit. Does this suffice?

lib/Migration/Version2300000Date20260720000001.php
```diff
@@ -0,0 +1,53 @@
...
+use OCP\Migration\Attributes\AddIndex;
+use OCP\Migration\Attributes\ColumnType;
+use OCP\Migration\Attributes\IndexType;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Introduces the `team_circle_id` column on `group_folders` so that the
+ * "this folder belongs to a team" relationship is stored as an explicit,
+ * indexed, queryable column.
```
artonge, line 25: That could not be done with the `goup_folders_groups` table?
Is `goup_folders_groups` still populated with the circle ID?

I fear that not populating `goup_folders_groups` could lead to subtle bugs as the code base currently assume that it is the source of truth.

I would rather add a column in the `group_folders` table to denote that a groupfolders assignment is limited to one group/circles only.

What do you think?

lib/TeamSpace/TeamSpaceProvider.php
artonge: Looks good :+1:

lib/Migration/Version2300000Date20260720000001.php
stediefan: Group_folders_groups is still populated via createTeamSpace() calling addApplicableGroup() (lib/TeamSpace/TeamSpaceService.php:58), so existing access and membership paths like getFoldersFromCircleMemberships() should continue to work without changes.

Team_circle_id is not a replacement for group_folders_groups. It indicates the ownership of the groupfolder in contrast to the Multi mounting prior (which is still possible with the non team-owned folders). Combining these in group_folders_groups.circle_id causes issues when admins assign multiple circles or re-add an owning circle. The unique index on team_circle_id also enforces "one team space per team" at the database level.

A boolean is_team_space column would still require a separate lookup to find the owning circle, whereas team_circle_id serves as both the flag and the reference in a single indexed read.

Both stores are synchronized: unlinkTeamSpace() clears team_circle_id and removes the group_folders_groups row (lib/TeamSpace/TeamSpaceService.php:122-123), maintaining group_folders_groups for all previous use cases and team_circle_id for the ownership.

PR #4976, opened 2026-07-31
fix(UI): fix ACL section subtitle font size

Jerome-Herbinet:
Current CSS font size variable sends "15px".

Other equivalent subtitles displayed above have "16px", not with a variable but simply with a hard-coded value.

Not tested (no test environment).

| Before (too small compared to others) | After (consistant size) | 
| --- | --- |
|  |  |

### DIFF
--- a/l10n/et_EE.js
+++ b/l10n/et_EE.js
@@ -60,6 +60,7 @@ OC.L10N.register(
     "Sort by number of groups that have access to this folder" : "Järjesta gruppide kaupa, kellel on ligipääs siia kausta",
     "Folder name" : "Kausta nimi",
     "Rename \"{mountPoint}\"" : "Muuda „{mountPoint}“ nime",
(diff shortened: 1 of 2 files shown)


kra-mo (changes requested):
I'd rather change the rest to use the proper variable, also for the font weight (`--font-weight-heading`)

src/components/SharingSidebarView.vue
```diff
@@ -536,6 +536,6 @@ thead .state-column {
 	margin-block: 4px 8px;
 	display: flex;
 	align-items: center;
-	font-size: var(--default-font-size);
+	font-size: var(--font-weight-heading);
```
kra-mo: This is setting a size to a weight, this will not work.

What I meant is to change the size of the others to also be the default font size instead of hardcoded.

kra-mo (changes requested):

Jerome-Herbinet:
Oops. Closed by error. I'll come back later. Sorry.

PR #5008, opened 2026-08-09
fix(versions): Fix getting the version mount

CarlSchwan:
Ensure we scan the folder if we don't find the entry in the cache.

## 🤖 AI (if applicable)


CarlSchwan:
Unlikely to fix the issue :(

PR #5011, opened 2026-08-10
feat: implement automatic creation of .system folder in auto-created team-folder

### LINKED ISSUE #2637: Store all team-related app-data in a team-folder

Fin-c:
feat: create an `.system` folder in the team-folder after creating a team. This hidden folder will hold team-owned app-data in the future.

Relates to https://github.com/nextcloud/circles/issues/2637

## 🤖 AI (if applicable)

- [x] The content of this PR was

### DIFF
--- a/lib/TeamSpace/TeamSpaceService.php
+++ b/lib/TeamSpace/TeamSpaceService.php
@@ -11,6 +11,8 @@ namespace OCA\GroupFolders\TeamSpace;
 
 use OCA\GroupFolders\Folder\FolderDefinition;
 use OCA\GroupFolders\Folder\FolderManager;
+use OCA\GroupFolders\Mount\FolderStorageManager;
+use OCP\Files\Storage\IStorage;
 use OCP\Teams\Team;
 use OCP\Teams\TeamFolder;
 use Psr\Log\LoggerInterface;
@@ -31,8 +33,11 @@ use Psr\Log\LoggerInterface;
  * needed.
  */
 class TeamSpaceService {
+	private const string APP_DIRECTORY_NAME = '.system';
+
 	public function __construct(
 		private readonly FolderManager $folderManager,
+		private readonly FolderStorageManager $folderStorageManager,
 		private readonly LoggerInterface $logger,
 	) {
 	}
@@ -51,6 +56,8 @@ class TeamSpaceService {
 		$folderId = $this->folderManager->createFolder($mountPoint);
 
 		try {
+			$this->createAppDirectory($folderId);
+
 			if ($quota > 0) {
 				$this->folderManager->setFolderQuota($folderId, $quota);
 			}
@@ -96,6 +103,34 @@ class TeamSpaceService {
 		}
 
 		return $folderId;
+	}
+
+	/**
+	 * Create the folder reserved for app data in a team space.
+	 *
+	 * @throws \RuntimeException when the folder cannot be created.
+	 */
+	private function createAppDirectory(int $folderId): void {
+		$teamSpace = $this->folderManager->getFolder($folderId);
+		if ($teamSpace === null) {
+			throw new \RuntimeException('Created team space could not be found');
+		}
+
+		$storage = $this->folderStorageManager->getBaseStorageForFolder(
+			$folderId,
+			$teamSpace->useSeparateStorage(),
+			$teamSpace,
+		);
+
+		if ($storage->is_dir(self::APP_DIRECTORY_NAME)) {
+			return;
+		}
+
+		if (!$storage->mkdir(self::APP_DIRECTORY_NAME) && !$storage->is_dir(self::APP_DIRECTORY_NAME)) {
+			throw new \RuntimeException('Could not create ' . self::APP_DIRECTORY_NAME . ' folder for team space');
+		}
+
+		$storage->getScanner()->scan(self::APP_DIRECTORY_NAME);
 	}
 
 	/**
@@ -196,6 +231,8 @@ class TeamSpaceService {
 		if ($folder === null) {
 			return null;
 		}
+
+		$this->createAppDirectory($folderId);
 
 		return new TeamFolder($folder->id, $folder->mountPoint);
 	}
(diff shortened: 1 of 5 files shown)


lib/TeamSpace/TeamSpaceService.php
```diff
@@ -98,6 +103,43 @@ public function createTeamSpace(string $circleId, string $mountPoint, int $quota
...
+			throw new \RuntimeException('Created team space could not be found');
+		}
+
+		$storage = $this->folderStorageManager->getBaseStorageForFolder(
+			$folderId,
+			$teamSpace->useSeparateStorage(),
+			$teamSpace,
+		);
+
+		$this->ensureSystemDirectory($storage, '.system');
```
mejo-: Isn't this meant to be `.apps`? Also, I would add this as a constant (e.g. `APP_FOLDER_NAME`) somewhere, possibly even in a public PHP class in server and reuse the constant everywhere instead of hardcoding the string in many places in code.

lib/TeamSpace/TeamSpaceService.php
```diff
@@ -98,6 +103,43 @@ public function createTeamSpace(string $circleId, string $mountPoint, int $quota
 		return $folderId;
 	}
 
+	/**
+	 * Create the folder reserved for system data in a team space.
+	 *
+	 * @throws \RuntimeException when the folder cannot be created.
+	 */
+	private function createSystemFolder(int $folderId): void {
```
mejo-: Function names should probably also be adjusted (`*AppFolder` instead of `*SystemFolder()`).

lib/TeamSpace/TeamSpaceService.php
Fin-c: I was told that it has been decided that the name if the folder is .system instead .apps.

I think that is a good idea. We have to enter the name in Server,right?

Fin-c: Or should I rename it to .app?

mejo-: Seems like the discussion for the folder name is ongoing (I'm in favor of `.apps`, but seems like `.team-app-data` is also in discussion). In any case, moving this to a constant will simplify to adjust it later. And I would call the function names `*AppDirectory()` as all names that are discussed at the moments are about "apps" in the name, not "system". Sorry for the confusion.

Regarding the home of the constant, I think you can keep it in `lib/TeamSpace/TeamSpaceService.php` for now and we can decide to move it to server code later on.

lib/TeamSpace/TeamSpaceService.php
Fin-c: Okay, thanks for the feedback. I've updated the function name as you described above, using AppDirectory.

lib/TeamSpace/TeamSpaceService.php
Fin-c: Don't worry, okay, it's done. I'd also guess that the constant would be better off in the “server” directory going forward. I've named the function ‘AppDirectory’ as you suggested, and for now I've named the constant folder “.system” until you guys can agree on a name.

susnux:
Keep this in mind: https://github.com/nextcloud/server/issues/46309

Meaning that folder must not have team specific contents in root level but should be considered to also contain other apps data to not create more folders like this in the future.

lib/TeamSpace/TeamSpaceService.php
```diff
@@ -98,6 +105,43 @@ public function createTeamSpace(string $circleId, string $mountPoint, int $quota
...
+	private function ensureAppDirectory(IStorage $storage, string $path): void {
+		if ($storage->is_dir($path)) {
+			return;
+		}
+
+		if (!$storage->mkdir($path) && !$storage->is_dir($path)) {
+			throw new \RuntimeException("Could not create $path folder for team space");
+		}
+
+		$storage->getScanner()->scan($path);
```
artonge: @icewind1991 is this the easiest way to go, or is there a better way?

lib/TeamSpace/TeamSpaceService.php
```diff
@@ -98,6 +105,43 @@ public function createTeamSpace(string $circleId, string $mountPoint, int $quota
...
+
+		$this->ensureAppDirectory($storage, self::APP_DIRECTORY_NAME);
+	}
+
+	/**
+	 * Ensure an internal app directory exists and is visible to the file cache.
+	 *
+	 * @throws \RuntimeException when the directory cannot be created.
+	 */
+	private function ensureAppDirectory(IStorage $storage, string $path): void {
```
artonge: Will this ever be used outside of `createAppDirectory`? If not, maybe we should merge both functions.

lib/TeamSpace/TeamSpaceService.php
Fin-c: @icewind1991 reminder

lib/TeamSpace/TeamSpaceService.php
Fin-c: i think not. we can merge

lib/TeamSpace/TeamSpaceService.php
artonge: He's in vacation, let's keep it as is and we can always improve later on.

artonge (approved):

PR #5071, opened 2026-08-31
Feature/implement default quota for groups

Fin-c:
implement Default Quota for TeamFolders 

issue:https://github.com/orgs/nextcloud/projects/150/views/1?pane=issue&itemId=229897063&issue=nextcloud%7Cgroupfolders%7C5029

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

PR #5075, opened 2026-09-03
rename team folder to group folder

### LINKED ISSUE #5063: Rename the former "Team folders" back to "Group folders"
### Task
In the effort to separate Team folders and Group folders, the latter should be uniformly renamed as such in the admin settings and everywhere else.

Fin-c:
Rename the Team folders back to group folders 
https://github.com/nextcloud/groupfolders/issues/5063

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


jospoortvliet (approved):
I looked through it, but didn't test (from what I can see it's all string changes).

* removing of team spaces is good move, we don't use the term
* all the changes in comments in the code - sure, maybe not that critical, but not a problem.
* warnings and such in UI - all look fine to me.

So 🐘  it looks good.

Minor thing - I think we should name it Group folders (capital, just like with Team). But to me, that's not a blocker, I rather get this in as-is than not at all.

tomek2k1:
Blocked by https://github.com/nextcloud/server/pull/63989 and https://github.com/nextcloud/groupfolders/pull/5076

Fin-c:
Blocked by https://github.com/nextcloud/groupfolders/pull/5071

lib/TeamSpace/TeamSpaceProvider.php
```diff
@@ -31,7 +31,7 @@ public function getId(): string {
 
 	#[\Override]
 	public function getName(): string {
-		return $this->l10n->t('Team spaces');
```
artonge, line 34: This will have to go in after the release for 35.0.1 as we cannot change translated strings that late in the process.
Please revert all of them and only keep non-translated strings

artonge (changes requested):

lib/TeamSpace/TeamSpaceProvider.php
Fin-c: I reverted all changes to translated strings and kept only the non-translated string changes

artonge (approved):

zak39:
Hi everyone! :)

Why did you rename this application again?

All users will get lost with these changes.

alimmroth:
> Hi everyone! :)
> 
> Why did you rename this application again?
> 
> All users will get lost with these changes.

Hello zak39, I know this can be confusing, but the decision was made to be able to bind team folders tighter to teams (auto-creating a team folder per team) and not using them for anything else while at the same time preserving the functionailty and structures that admins have build - under the name group folders. This makes the use case for both of them clearer.

